PR #6, opened 2020-02-24
user authentication design and workflow

aquibbaig:
This is the Initial PR for User authentication design.


anishaswain:
Getting tracked at #42

PR #7, opened 2020-02-26
pbench-result-data-sample doc type migration (phase one)

gurbirkalsi:
Phase one of result-data index migration represents changes to
postprocessing when handling run iteration & sample data. Data from the new index is now being processed and visualized within the `ComparisonSelect`, `RunComparison,` and `Summary` page components. 

Queries are made to the new Elasticsearch index whereas previously queries were made to
result.json files on the production results server.

Phase two will represent changes to timeseries data retrieval, visualization, and UX updates to cluster legends, tables, and percent change graphs using Patternfly. 

Notable changes:
- `result-index` key in dashboard config references new index
- postprocessing of iteration & sample data now occurs within redux action flow
- implementation of cluster generation using ES6 object spread & object literals

### DIFF
--- a/src/services/dashboard.js
+++ b/src/services/dashboard.js
@@ -1,12 +1,14 @@
-import _ from 'lodash';
 import request from '../utils/request';
-import { renameProp } from '../utils/utils';
 
-function parseMonths(datastoreConfig, selectedIndices) {
+function parseMonths(datastoreConfig, index, selectedIndices) {
   let indices = '';
 
   selectedIndices.forEach(value => {
-    indices += `${datastoreConfig.prefix + datastoreConfig.run_index + value},`;
+    if (index === datastoreConfig.result_index) {
+      indices += `${datastoreConfig.prefix + index + value}-*,`;
+    } else {
+      indices += `${datastoreConfig.prefix + index + value},`;
+    }
   });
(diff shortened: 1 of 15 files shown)


src/services/dashboard.js
```diff
@@ -111,12 +116,147 @@ export async function queryTocResult(params) {
...
+                field: 'run.name',
+              },
+            },
+            controller: {
+              terms: {
+                field: 'run.controller',
+              },
+            },
+          },
+          size: 10000,
```
gurbirkalsi, line 193: Phase two will include a smaller scroll size and utilize Promises to continuously aggregate data on the open scroll query until we identify an empty object.

portante (approved):
Looks good, but only from a cursory view.  I don't pretend to understand all the details since there some subtly here.

Having a way to mock the Elasticsearch queries so that we can see what the ES response looks like and then how the code parses that response into a display would be helpful.

gurbirkalsi:
@portante I've added unit tests for verifying the post processing steps defined in `parse.js` output the expected mocked data.

portante (approved):
Looks good, now we have to verify it in a staging environment.

anishaswain (approved):

PR #12, opened 2020-03-17
Update indices query with faster call for concise list of all indices

gurbirkalsi:


src/services/datastore.js
```diff
@@ -29,7 +29,7 @@ export async function queryDatastoreConfig() {
 export async function queryMonthIndices(params) {
   const { datastoreConfig } = params;
 
-  const endpoint = `${datastoreConfig.elasticsearch}/_cat/indices?format=json&pretty=true`;
+  const endpoint = `${datastoreConfig.elasticsearch}/_aliases`;
```
portante, line 32: Does `_aliases` give you JSON output?

portante (changes requested):

src/services/datastore.js
gurbirkalsi: Yes, it returns JSON with pretty formatting by default.

portante: But our production V1 Elasticsearch instance returns an empty list because we don't use aliases.

portante: Okay, I can see how this might be more efficient, as it does return JSON and is a much smaller payload.

portante (approved):
Looks good.

PR #42, opened 2020-04-27
User authentication mocked flow with PatternFly

anishaswain:
The PR intends to mock the workflow of the user authentication process (exp. Login, Signup, Forgot password) by imitating the user notion.
The wireframe of the implementation:[ wireframe](https://marvelapp.com/dh1cd91/screen/66676164)
Page View:

1. Auth Page

Screenshot

2. Login Page 

Screenshot (1)

3. Signup Page
Screenshot (1)

4. Forgot Password Page

Screenshot


anishaswain:
@portante @gurbirkalsi The mock implementation is ready for review now. Apart from the SSO, GitHub and Gmail login, all the other pages (login, signup, forgot password) are working. As of now, the hardcoded username and password are both "admin" for login purpose. Let me know if we also need to implement mock signup feature. 

Also, the end-to-end test for all the pages of the dashboard are failing now in travis. (Not sure why?)

src/components/AuthLayout/index.js
```diff
@@ -0,0 +1,162 @@
...
+      <div>
+        <Title headingLevel="h2" size="3xl" className={styles.section}>
+          {backOpt === 'true' ? back : <Fragment />}
+          <br />
+          {heading}
+          <Dropdown
+            style={{ float: 'right' }}
+            onSelect={this.onSelect}
+            toggle={
+              <DropdownToggle
```
gurbirkalsi, line 90: I'm receiving the following error when opening this DropdownToggle: 
`Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.`

src/components/AuthLayout/index.js
```diff
@@ -0,0 +1,162 @@
...
+            mdOffset={6}
+            lgOffset={6}
+            className={styles.sideGrid}
+          >
+            <div>
+              <img src={logo} alt="pbench_logo" className={styles.logo} />
+            </div>
+            <div className={styles.sideGridItem}>
+              <Title headingLevel="h4" size="xl">
+                PBench is a harness that allows data collection from a variety of tools while
```
gurbirkalsi: Only "P" is capital in Pbench

src/pages/PrivateRoute/index.js
```diff
@@ -0,0 +1,121 @@
...
+      <Grid gutter="md" className={styles.restLoginDiv}>
+        <GridItem>
+          <Title headingLevel="h4" size="xl">
+            Need an account?
+            <Button
+              variant="link"
+              onClick={() => this.navigate('signup')}
+              className={styles.inlineLink}
+            >
+              signup
```
gurbirkalsi: `Sign up` should be two words and have a capital `S`. We should also consider adding padding to the left of this button as it is currently rendering right next to the `Title` parent element.

src/pages/PrivateRoute/index.js
```diff
@@ -0,0 +1,121 @@
...
+        {PbenchMethods}
+        {restLoginHandlers}
+        {loginMethods}
+      </Fragment>
+    );
+
+    if (user.user.username) {
+      return <Fragment>{children}</Fragment>;
+    }
+    return <AuthLayout toPreview={toPreview} heading="Login With .." />;
```
gurbirkalsi: The heading text should be `Log in with ...` according to the mockups

src/pages/PrivateRoute/index.js
```diff
@@ -0,0 +1,121 @@
...
+            className={styles.btn}
+          >
+            <p className={styles.inlineLink}>Gmail</p>
+          </Button>
+        </GridItem>
+      </Grid>
+    );
+    const toPreview = (
+      <Fragment>
+        {PbenchMethods}
```
gurbirkalsi: Can we consider renaming `PbenchMethod` to `pbenchLogin`, `restLoginHandlers` to `authForms`, and `loginMethods` to `thirdPartyLogin` for more clarity into what is being rendered?

src/pages/SignupHandler/index.js
```diff
@@ -0,0 +1,267 @@
...
+      password: '',
+      confPassword: '',
+      email: '',
+      variantVal: 'tertiary',
+      btnColor: 'black',
+    };
+  }
+
+  componentDidMount = () => {
+    this.disableSubmitBtn();
```
gurbirkalsi, line 42: What if we provided visual feedback to the input forms instead of disabling the button upon component mount when the user has not filled out the sign up form? We should also validate the user has entered the correct field (for example, verify user has entered an email address in the email address field). Keep in mind, this validation may be moved to the server in the future.

src/pages/SignupHandler/index.js
```diff
@@ -0,0 +1,267 @@
...
+  };
+
+  enableSubmitBtn = () => {
+    this.setState({
+      variantVal: 'primary',
+      btnColor: 'white',
+    });
+    const btn = document.getElementById('submitBtn');
```
gurbirkalsi: We should use React functionality here as opposed to the DOM API. Patternfly's Button allows us to accomplish this using the API it exposes.

src/pages/SignupHandler/index.js
```diff
@@ -0,0 +1,267 @@
...
+            isRequired
+            type="password"
+            id="horizontal-form-confirm-password"
+            name="horizontal-form-confirm-password"
+            onChange={this.checkConfirmPassword}
+          />
+        </FormGroup>
+        <p style={{ color: 'red' }} id="passwordStatus" />
+        <div id="passworkCheck">
+          <h1>Password must contain atlest</h1>
```
gurbirkalsi: Should be `at least` here

src/pages/SignupHandler/index.js
```diff
@@ -0,0 +1,267 @@
...
+          </Split>
+          <Split id="upperCaseChar">
+            <SplitItem style={{ marginTop: '2px' }}>
+              <CloseIcon style={{ color: 'red' }} />
+            </SplitItem>
+            <SplitItem style={{ marginTop: '2px', display: 'none' }}>
+              <CheckIcon style={{ color: 'green' }} />
+            </SplitItem>
+            <SplitItem isFilled style={{ marginLeft: '15px' }}>
+              1 upercase letter
```
gurbirkalsi: Should be `uppercase` here

src/utils/authority.js
```diff
@@ -1,7 +1,7 @@
 // use localStorage to store the authority info, which might be sent from server in actual project.
 export function getAuthority() {
   // return localStorage.getItem('antd-pro-authority') || ['admin', 'user'];
-  return localStorage.getItem('antd-pro-authority') || 'admin';
+  return localStorage.getItem('antd-pro-authority') || 'admin1';
```
gurbirkalsi: We can remove `authority.js` in addition to the `Authorized.js` definition as this is not being used for managing users.

src/models/user.js
```diff
@@ -4,9 +4,17 @@ export default {
   state: {
     favoriteControllers: [],
     favoriteResults: [],
+    user: {},
```
gurbirkalsi: We should avoid naming state variables that have the same names as the model namespace in order to improve our store's organization.

src/models/user.js
```diff
@@ -4,9 +4,17 @@ export default {
...
   },
 
   effects: {
+    *loadUser({ payload }, { put }) {
+      console.log(payload);
```
gurbirkalsi: Our linter should have caught this console.log() statement here. Do we need this here?

src/app.js
```diff
@@ -14,7 +14,7 @@ const persistConfig = {
   throttle: 1000,
   key: getAppPath(),
   storage,
-  whitelist: ['global', 'user'],
```
gurbirkalsi, line 18: We should consider preserving the `user` namespace here as the authentication flow involves caching user credentials and then eventually confirming their validity against the server.

src/components/AuthLayout/index.js
anishaswain: fixed

src/components/AuthLayout/index.js
anishaswain: Fixed

src/pages/PrivateRoute/index.js
anishaswain: fixed

src/pages/PrivateRoute/index.js
anishaswain: fixed

src/pages/PrivateRoute/index.js
anishaswain: Renamed

src/pages/SignupHandler/index.js
anishaswain: Yes, we can do that as well. We should discuss these design constraints with Mary. But as per our last discussion, these constraints are not our top priority right now as it might change in future so this will need further discussion.

src/pages/SignupHandler/index.js
anishaswain: Resolved

src/utils/authority.js
anishaswain: @gurbirkalsi when I removed these two files, `BasicLayout.js` was throwing error as it is referenced in many places there. Do we want to remove the references from there as well?

src/models/user.js
anishaswain: solved!

src/app.js
anishaswain: @gurbirkalsi if we preserve the `user` namespace, the state of redux for user will not getting cleared even after refresh and it is automatically taking up the value from the previous login. That is, if the user doesn't login and go back to the PrivateRoute, it will still take the username of the previously logged in user. Is it an expected behaviour for log in system?

src/pages/SignupHandler/index.js
gurbirkalsi: We can expect this sign up form design to be constant for the foreseeable future. At the very least, we might add or remove a field; however, our component markup should be dynamic enough that this wouldn't alter core logic.

src/utils/authority.js
gurbirkalsi: Yes, lets remove the references throughout the codebase!

src/app.js
gurbirkalsi: We should have a `logout` Redux action that takes the `user` namespace and clears all authentication related fields from the namespace state. Eventually, we will confirm the validity of the cached user token against the server which will then be deleted or preserved (via Redux actions) depending on the result of the confirmation.

In addition, we should consider creating an `auth` namespace as the `user` namespace houses global state for user preferences/favorites that we should avoid mixing with `auth`.

gurbirkalsi:
@Anisha1234 It'd be great to also include unit, snapshot, and e2e tests as part of this work!

src/components/AuthLayout/index.js
```diff
@@ -0,0 +1,175 @@
...
+            lgOffset={6}
+            className={styles.sideGrid}
+          >
+            <div>
+              <img src={logo} alt="pbench_logo" className={styles.logo} />
+            </div>
+            <div className={styles.sideGridItem}>
+              <Title headingLevel="h4" size="xl">
+                Pbench is a harness that allows data collection from a variety of tools while
+                running a benchmark. PBench has some built-in script that run some common
```
gurbirkalsi: `B` in Pbench is still capitalized here

anishaswain:
> @Anisha1234 It'd be great to also include unit, snapshot, and e2e tests as part of this work!

On it!

aquibbaig (changes requested):
Just nitpicking a few things, Rather than using "document" object to modify the DOM tree, React has a pre-defined way of using Virtual DOM to be manipulated using it's components. So, in most cases(repositories) that use React, we do not see the the use of `document.getElementById()` to modify the DOM elements. We need to have a "single source of truth" while manipulating our DOM, else we can run into trouble in the future.
If we seriously have a dire need to do that, then we can use `refs` as suggested in the following articles.
- http://jamesknelson.com/react-js-by-example-interacting-with-the-dom/
- https://stackoverflow.com/questions/38093760/how-to-access-a-dom-element-in-react-what-is-the-equilvalent-of-document-getele

The best way to solve our requirement is by using a component that interacts with the virtual DOM by default. For example, each sign up form field would have a validate key that actually takes in an array of errors to which we can append each time the state of the current field changes. So, after user has finished filling up the form, if there are no errors, a current field is "sanitised".

anishaswain:
> Just nitpicking a few things, Rather than using "document" object to modify the DOM tree, React has a pre-defined way of using Virtual DOM to be manipulated using it's components. 

Yes, you are right. I have changed the direct DOM manipulations.

src/pages/SignupHandler/index.js
anishaswain: Added visual feedback for email and password field

src/app.js
anishaswain: I have created different namespaces for different auth and user data.

src/e2e/controllers.e2e.js
```diff
@@ -11,6 +11,26 @@ beforeAll(async () => {
   });
   page = await browser.newPage();
   await page.goto('http://localhost:8000/dashboard/#/');
+  // Login using dummy credentials
```
gurbirkalsi, line 15: We need to move these e2e tests to the `describe()` section as individual tests below as the `beforeAll()` section only handles initialization of the test environment.

anishaswain: @gurbirkalsi there was issue with the submit button id in each test file and in `search.e2e.js`, it wasn't getting redirected to search page. I have fixed both of them.

gurbirkalsi:
@Anisha1234 Can you please squash these down and rebase on master?

anishaswain:
> @Anisha1234 Can you please squash these down and rebase on master?

@gurbirkalsi done

gurbirkalsi:
@Anisha1234 It seems the rebase broke component references in the `Profile` page component

gurbirkalsi (approved):
Great work @Anisha1234 !

PR #41, opened 2020-03-25
Phase two result-data-sample migration

gurbirkalsi:
See also #7.

PR #49, opened 2020-05-13
Install and setup patternfly

aquibbaig:
This PR  installs and sets up patternfly for the dashboard.

## Motivation
Since we are moving towards "patternfly" for all our UI components, it was necessary to get it installed first rather than trying out all those features that would require patternfly as a dependency to be installed first. This will keep other people unblocked of their works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
     "@antv/g2": "^3.4.10",
     "@babel/runtime": "^7.3.1",
     "@nivo/bar": "^0.36.0",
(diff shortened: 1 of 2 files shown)


gurbirkalsi:
PR #41 provides the base installation for PatternFly dependencies along with CSS configuration. Can we consider dropping this PR in favor for #41?

aquibbaig:
Thanks @gurbirkalsi. Closing this one

PR #50, opened 2020-05-15
Initial design for the DatePicker component

### LINKED ISSUE #46: enhanced time range picker component
Currently the month picker component only allows users to select ranges of data by months. In most cases, users are concerned with the last week's set of data. In order to provide further granularity to time range selection, we will build a component to allow users to set relative and absolute time ranges. 

Example design: https://dribbble.com/shots/4433950-Datepicker-with-Time-Range-Slider

aquibbaig:
Fixes #46.

This will be a In-progress PR for now until I work on the following changes:
- [x] Date picker calendar range selector.
- [x]  Enable aggregation selections like "Last month", "This month", etc. over the calendar.
- [ ]  Adhere to Patternfly design principles (Important).
- [ ]  Write tests to ensure that the component is working properly.
- [ ]  Remove MonthSelect component and ensure DatePicker does the same functionality.
- [ ]  Media queries?

### DIFF
--- a/src/components/Calendar/index.js
+++ b/src/components/Calendar/index.js
@@ -1,0 +1,204 @@
+import React, { Component } from 'react';
+import moment from 'moment';
+
+import styles from './index.less';
+
+// Calendar component is the actual dropdown
+// Calendar that picks a range of dates.
+// TODO: Add aggregated date picks such
+// as "This week", "This month", etc.
+export default class Calendar extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      dateObject: moment(),
+      startDate: moment(),
+      endDate: moment(),
+    };
+    this.changeMonth = this.changeMonth.bind(this);
+    this.resetDate = this.resetDate.bind(this);
+    this.changeDate = this.changeDate.bind(this);
+  }
+
+  // Sets the state of the parent component
+  // when it updates it's state.
+  componentDidUpdate(prevProps, prevState) {
+    const { startDate, endDate } = this.state;
+    const { updateDate, updateDropDown } = this.props;
+    // Update the parent component.
+    if (startDate !== prevState.startDate) {
+      updateDate(startDate, endDate);
+      updateDropDown(startDate, endDate);
+    }
+    if (endDate !== prevState.endDate) {
+      updateDate(startDate, endDate);
+      updateDropDown(startDate, endDate);
+    }
+  }
+
+  // Changes the month selection.
+  changeMonth = month => {
+    const { dateObject } = this.state;
+    const newDateObject = dateObject.month(month);
+    this.setState({
+      dateObject: newDateObject,
+    });
+  };
+
+  // Resets the dateObject.
+  resetDate = () => {
+    this.setState({ dateObject: moment() });
+  };
+
+  // When we change the date. (startDate and endDate)
+  changeDate = date => {
+    let { startDate, endDate } = this.state;
+    if (
+      startDate === null ||
+      date.isBefore(startDate, 'day') ||
+      !startDate.isSame(endDate, 'day')
+    ) {
+      startDate = moment(date);
+      endDate = moment(date);
+    } else if (date.isSame(startDate, 'day') && date.isSame(endDate, 'day')) {
+      startDate = null;
+      endDate = null;
+    } else if (date.isAfter(startDate, 'day')) {
+      endDate = moment(date);
+    }
+    this.setState({
+      startDate,
+      endDate,
+    });
+  };
+
+  render() {
+    const { dateObject, startDate, endDate } = this.state;
+    // MonthNavigation component is for choosing calendar months.
+    const MonthNavigation = ({ date, changeMonth, resetDate }) => {
+      return (
+        <nav className={styles.calendarNav}>
+          <a onClick={() => changeMonth(date.month() - 1)}>&#8249;</a>
+          <h1 onClick={() => resetDate()}>
+            {date.format('MMMM')} <small>{date.format('YYYY')}</small>
+          </h1>
+          <a onClick={() => changeMonth(date.month() + 1)}>&#8250;</a>
+        </nav>
+      );
+    };
+    // MonthDays component represents each month in the calendar.
+    const MonthDays = ({ date, startOfTheMonth, endOfTheMonth, onClick }) => {
+      const thisDate = moment(date);
+      const daysInMonth = moment(date).daysInMonth();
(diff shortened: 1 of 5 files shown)


portante:
@aquibbaig, converted to a "Draft" instead of using the "WIP" label, since that method was recently added by GitHub and is very useful.

gurbirkalsi:
@aquibbaig Did the theming support for https://github.com/Adphorus/react-date-range and https://github.com/airbnb/react-dates allow us to customize the component with PatternFly colors? If so, using the libraries would be the preferred option here as writing a custom date picker component is going to require more development overhead and maintenance. Let me know your thoughts!

aquibbaig:
@gurbirkalsi I tried using `airbnb/date-picker` but I can't seem to override their themes as installing the default interface throws a "MODULE_NOT_FOUND" error. Also, overriding default css won't work. Other than that, `reac-date-range` has peer dependencies on their css imports. Need some discussion before moving forward. The custom solution would work independent of all global css and theme overrides.

anishaswain:
@aquibbaig Could you please provide me with some insights on what styles you want to change over there with `airbnb/date-picker`?

aquibbaig:
> @aquibbaig Could you please provide me with some insights on what styles you want to change over there with `airbnb/date-picker`?

https://www.patternfly.org/v4/documentation/overview/global-css-variables/

gurbirkalsi:
I was able to override default css for `react-date-range` by defining a custom css file which references their component class names. Feel free to play around with the `styles.css` definition at this codesandbox environment: https://codesandbox.io/s/dark-sun-tt2rq

I've included the default global PatternFly css import which is present in the dashboard. This gives us the typography we are looking for. If we can get PatternFly primary colors, shadows, and layout behavior working we should be in good shape to consume this library: https://www.patternfly.org/v4/design-guidelines/styles/colors#primary-colors

anishaswain:
@aquibbaig @gurbirkalsi the default CSS for  `airbnb/date-picker` can also be overridden (for exp.  here [https://codesandbox.io/s/react-dates-date-range-picker-ti60b](https://codesandbox.io/s/react-dates-date-range-picker-ti60b  )). But when I tried to implement the same in dashboard, the `index.less` file wasn't getting loaded. Any idea why that might be happening @gurbirkalsi ? Do we need to use css-loader here or something?

Also `airbnb/date-picker` uses another npm package named [react-with-styles](https://github.com/airbnb/react-with-styles/blob/master/README.md#interfaces) for theming. mentioned [here](https://github.com/airbnb/react-dates#theming). So if we can customize `react-date-range`  then I guess we should go with that only.

gurbirkalsi:
We're loading `.less` files in `global.js` and don't seem to have any problems there. Was there a specific error you received?

aquibbaig:
@gurbirkalsi I also faced the same thing. I couldn't seem to override the styles from `airbnb/react-dates` because we probably use webpack to compile assets on load. They use theming from interfaces. And `react-date-range` also uses other packages to style itself. I can have both of the implementations and we chose from either one.

aquibbaig:
> @aquibbaig @gurbirkalsi the default CSS for `airbnb/date-picker` can also be overridden (for exp. here https://codesandbox.io/s/react-dates-date-range-picker-ti60b). But when I tried to implement the same in dashboard, the `index.less` file wasn't getting loaded. Any idea why that might be happening @gurbirkalsi ? Do we need to use css-loader here or something?
> 
> Also `airbnb/date-picker` uses another npm package named [react-with-styles](https://github.com/airbnb/react-with-styles/blob/master/README.md#interfaces) for theming. mentioned [here](https://github.com/airbnb/react-dates#theming). So if we can customize `react-date-range` then I guess we should go with that only.

@Anisha1234 The sandbox doesn't override all the elements, as mentioned in the documentation, we need to register a theme (e.g. aphrodite) and then probably override the variables.

aquibbaig:
> I was able to override default css for `react-date-range` by defining a custom css file which references their component class names. Feel free to play around with the `styles.css` definition at this codesandbox environment: https://codesandbox.io/s/dark-sun-tt2rq
> 
> I've included the default global PatternFly css import which is present in the dashboard. This gives us the typography we are looking for. If we can get PatternFly primary colors, shadows, and layout behavior working we should be in good shape to consume this library: https://www.patternfly.org/v4/design-guidelines/styles/colors#primary-colors

@gurbirkalsi Don't you think the component offered by `react-dates` is too big for our page?

aquibbaig:
@gurbirkalsi @Anisha1234 This is what this PR is about. This custom component was easier to prepare and designed as per patternfly standards. It also uses less space than other available libraries out there. Also, we can configure it easily according to the needs.

Custom-date-picker

anishaswain (changes requested):
@aquibbaig There are some points that we might need to cover

1. The custom design seems to be inconsistent with both patternfly and the current dashboard design. (The component is almost taking 60% of the page space which can be reduced by reducing padding between dates and font size etc.)
2. The range selector(feature of highlighting the dates) is not working if we are selecting the aggregation selections like "_Last month_", "_This month_".
3. The range selector is taking the today's date instead of "last day", which I believe should be the date of yesterday. (otherwise also we need to be incisive about selecting yesterday's and today's data)
4. We might like to have more aggregation selections like "15 days" (Not sure about it)
5. The dropdown option seems to be little off for aggregation selections. Is it possible to somehow add it in a side view which will give a better user experience I believe. In this way we can clearly draw a distinction between custom selection and aggregation selection as well.

aquibbaig:
> @aquibbaig There are some points that we might need to cover
> 
> 
> 
> 1. The custom design seems to be inconsistent with both patternfly and the current dashboard design. (The component is almost taking 60% of the page space which can be reduced by reducing padding between dates and font size etc.)
> 
> 2. The range selector(feature of highlighting the dates) is not working if we are selecting the aggregation selections like "_Last month_", "_This month_".
> 
> 3. The range selector is taking the today's date instead of "last day", which I believe should be the date of yesterday. (otherwise also we need to be incisive about selecting yesterday's and today's data)
> 
> 4. We might like to have more aggregation selections like "15 days" (Not sure about it)
> 
> 5. The dropdown option seems to be little off for aggregation selections. Is it possible to somehow add it in a side view which will give a better user experience I believe. In this way we can clearly draw a distinction between custom selection and aggregation selection as well.

I agree to the above points and I think we can move the aggregation selections inside the calendar itself, and also use them to highlight the range, it’s an easy step. Also regarding the size, we need to reduce the font and padding.

aquibbaig:
Closing this PR on account of recent discussion!

PR #51, opened 2020-05-18
Fixes links in footer

anishaswain:
Adds Dashboard, Dashboard Github repo and Pbench Github repo link in footer.

### DIFF
--- a/src/layouts/BasicLayout.js
+++ b/src/layouts/BasicLayout.js
@@ -278,15 +278,6 @@ class BasicLayout extends React.PureComponent {
             <GlobalFooter
               links={[
                 {
-                  key: 'Pbench Dashboard',
-                  title: 'Pbench Dashboard',
-                  href:
-                    process.env.NODE_ENV === 'development'
-                      ? `http://${window.location.hostname}:${window.location.port}`
-                      : `${datastoreConfig.results}/dashboard`,
-                  blankTarget: true,
-                },
-                {
                   key: 'github',
                   title: <Icon type="github" />,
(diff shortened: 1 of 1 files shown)


src/layouts/BasicLayout.js
```diff
@@ -280,19 +280,19 @@ class BasicLayout extends React.PureComponent {
                 {
                   key: 'Pbench Dashboard',
                   title: 'Pbench Dashboard',
-                  href: 'https://example.com',
+                  href: 'http://pbench.perf.lab.eng.bos.redhat.com/dashboard/',
```
portante: Isn't this one is relative to the deployment?  If so we need to use the loaded configuration to construct the URL.  Since this should be a different URL for production, vs. staging, vs. local laptop deployment.

portante (changes requested):

portante (approved):

gurbirkalsi (approved):

src/layouts/BasicLayout.js
```diff
@@ -280,19 +280,22 @@ class BasicLayout extends React.PureComponent {
                 {
                   key: 'Pbench Dashboard',
                   title: 'Pbench Dashboard',
-                  href: 'https://example.com',
+                  href:
```
gurbirkalsi: `datastoreConfig.results` would give us an endpoint of `/results/dashboard` here which would route the user to an unknown page. We can consider removing this key altogether as the user is already at the Pbench Dashboard link.

anishaswain: fixed

gurbirkalsi (approved):

PR #77, opened 2020-08-04
User profile and account settings page component

### LINKED ISSUE #47: user account settings page component
User authentication will require a settings page for updating common user details such as email, password, etc. In order to provide this functionality, we should mock a user account settings page component with a dedicated route within the pbench dashboard config. The mocked page component should utilize the namespace created for handling authentication for modifying user details.

anishaswain:
Fixed #47 
User authentication requires a settings page for updating common user details such as email, password, etc.  The user profile page also intends to show different types of data related or owned by the use.

1.Mocked Design:
Screen Shot 2020-08-06 at 2 12 00 PM

2. Implemented version:

3. Edit View

4. Reset Password


src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
     this.state = {
-      isChecked: true,
+      dataExpireValue: '6 months',
+      editView: false,
+      displayName: 'Anisha Swain',
+      emailVal: 'xyz.redhat@gmail.com',
+      positionVal: 'Associate Software Engineer',
+      accountType: 'Pbench',
+      accountStatus: 'Active',
+      color: '#0080005e',
```
dbutenhof: Placeholder data in the source tree?

src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
+                                <Text component={TextVariants.h5}>RH329821</Text>
+                              </TextContent>
+                            ) : (
+                              <TextInput
+                                value="RH329821"
+                                type="text"
+                                isReadOnly
+                                aria-label="RH329821"
+                                className={styles.editInput}
+                              />
```
dbutenhof: Why is ID not in the state?

src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
+                        </Grid>
+                        <Grid>
+                          <GridItem span={12} className={styles.subCardDiv}>
+                            <Button variant="link" icon={<RedoIcon />}>
+                              Reset All Data
+                            </Button>{' '}
+                          </GridItem>
+                          <GridItem span={12} className={styles.subCardDiv}>
+                            <TextContent>
+                              <Text component={TextVariants.h5}>No of results expiring</Text>
```
dbutenhof: I'd rather see "Number" than "No" ... Or maybe just reverse it and have "123 results expiring".

The hardcoded "123" seems odd here, too, though at this point it'd be hardcoded somewhere, like in a reducer.

dbutenhof (approved):
As a mockup, this mostly seems OK, although the hardcoded placeholders bother me a bit.

src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
+      color: '#0080005e',
     };
-  }
 
-  componentDidMount = () => {};
+    this.dataExpireOptions = [
+      { value: '2 months', label: '2 months', disabled: false },
+      { value: '6 months', label: '6 months', disabled: false },
+      { value: '1 year', label: '1 year', disabled: false },
+    ];
```
npalaska: Shouldnt one of these values be enabled serving as a default (ideally placeholder value)

src/pages/Profile/index.js
aquibbaig: might consider getting this data from an API request, or in worst case a `mock` file

src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
+
+    this.accountOptions = [
+      { value: 'Pbench', label: 'Pbench', disabled: false },
+      { value: 'Google', label: 'Google', disabled: false },
+      { value: 'RedHat SSO', label: 'RedHat SSO', disabled: false },
+      { value: 'Github', label: 'Github', disabled: false },
```
aquibbaig: we can use a `mock` file for that

src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
+      this.setState({
+        editView: !editView,
+      });
+    };
+
+    this.saveEdit = () => {
+      const { editView, displayName, emailVal, positionVal } = this.state;
+      const { name, email, position } = this.ref;
+
+      const editedName = name.current.value === '' ? displayName : name.current.value;
```
aquibbaig: can be changed to:
```
name.current.value ? displayName : name.current.value;
```

src/pages/Profile/index.js
```diff
@@ -23,90 +32,361 @@ import avatar from '../../assets/avatar.svg';
...
+      const editedEmail = email.current.value === '' ? emailVal : email.current.value;
+      const editedPosition = position.current.value === '' ? positionVal : position.current.value;
+
+      this.setState({
+        editView: !editView,
```
aquibbaig: `editView: !editView,` => `!editView`

aquibbaig (changes requested):
Great work @anishaswain. Just made some small changes on the PR. I think we could make use of a mock file named `user1.json` to fetch our mock user content to populate the state on a `useEffect` after first render. Also, we can try leveraging a functional component and hooks now, since we have support for that now, after upgrading to React 16.8.

src/pages/Profile/index.js
anishaswain: Need to put the hardcoded value in the mock file. Converting the PR into the draft for the same.

anishaswain:
Need to put the hardcoded value in the mock file. Converting the PR into a draft for the same.

src/pages/Profile/index.js
anishaswain: I thought the ID of an employee won't change. Should I make it editable?

src/pages/Profile/index.js
anishaswain: fixed!

src/pages/Profile/index.js
anishaswain: Fixed! Added to mock file.

src/pages/Profile/index.js
anishaswain: Yes, the default value is set in the `component state` which is grabbed from a mock user file(_which will be replaced with data from the server later_)

The `dataExpireOptions` array is just from showing the values in the dropdown in a proper format.

src/pages/Profile/index.js
anishaswain: The `accountOptions` array is just for showing the values in the dropdown in a proper format. Plus I don't think we are going to store all the options as user data in the backend. Are we!

However, the type of the user account needs to be stored in backend, so it is set in the component state which is grabbed from a mock user file(which will be replaced with data from an API request later)

src/pages/Profile/index.js
anishaswain: Added `mock` file

src/pages/Profile/index.js
anishaswain: changed!

src/pages/Profile/index.js
anishaswain: `Parsing error: Unexpected token`

src/pages/Profile/index.js
gurbirkalsi: +1,  if ID is a read-only property we should consider referencing it as a prop.

anishaswain: @gurbirkalsi @dbutenhof added the ID in the mock file as a prop. Removed the reference from the state.

dbutenhof: I don't know whether "ID" is or should be static; is the meaning strongly defined in the mocks anywhere? It's not even necessarily true that an employee ID doesn't change over time; for example when Compaq absorbed Digital, the older low numbered Digital badge numbers (including mine) overlapped and got changed. It would have been mildly annoying to be using a database system like this that didn't allow correcting that.

mock/user.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable import/prefer-default-export */
+export const profileData = {
+  dataExpireValue: '6 months',
+  editView: false,
+  displayName: 'Anisha Swain',
+  emailVal: 'xyz.redhat@gmail.com',
+  positionVal: 'Associate Software Engineer',
+  accountType: 'Pbench',
+  accountStatus: 'Active',
+  resultsExpiring: '213',
```
dbutenhof: Shouldn't companyID be here?

src/pages/Profile/index.js
```diff
@@ -38,13 +39,13 @@ class Profile extends React.Component {
...
-      displayName: 'Anisha Swain',
-      emailVal: 'xyz.redhat@gmail.com',
-      positionVal: 'Associate Software Engineer',
-      accountType: 'Pbench',
-      accountStatus: 'Active',
+      dataExpireValue: profileData.dataExpireOptions,
+      editView: profileData.editView,
+      displayName: profileData.displayName,
+      emailVal: profileData.emailVal,
+      positionVal: profileData.positionVal,
```
dbutenhof, line 50: And why wouldn't companyID be in the state?

dbutenhof (commented):
I had a sudden thought while reading this again that it's a bit odd that "results expiring" is part of the user profile. It's not a static "user identification" property that will be fetched from the GraphQL DB with username/password/etc ... it's a dynamic query on the Elasticsearch for data sets owned by the current user and within some timestamp range. (And, depending on the query ranges, it could theoretically change at any moment.)

Seems it'd be useful to have a warning on the user's dashboard, perhaps as a dismissable warning banner; but it's an entirely different kind of information from the "profile" for the user. I'm curious why it was designed this way?

src/pages/Profile/index.js
```diff
@@ -23,90 +33,366 @@ import avatar from '../../assets/avatar.svg';
...
   constructor(props) {
     super(props);
+    this.ref = {
+      name: React.createRef(),
+      email: React.createRef(),
+      position: React.createRef(),
+    };
     this.state = {
-      isChecked: true,
+      dataExpireValue: profileData.dataExpireOptions,
```
dbutenhof: I didn't notice this until re-reading Nikhil's comment below, but the mock data property is `dataExpireValue` and the property referenced here doesn't appear to exist in the `mock/user.js` file's `profileData`.

Also, it still seems to me that we should be loading the mock `companyID` into the state, even if we don't allow editing it; the "real code" won't be able to see it otherwise without a `REST` call for every render.

src/pages/Profile/index.js
dbutenhof: Come to think of it ... why are we hardcoding the data retention intervals in the React component? Eventually the backend will need to be processing and validating these. I don't know how these came to be in the original design and UI mocks, but I wonder if these should be "enshrined" in Redux with a REST mock file like the user profile?

src/pages/Profile/index.js
dbutenhof: My assumption is that we use SSO to authenticate a user and provide initial registration information, but subsequently they'd log in using a Pbench username/email and password. In that case, it may be that the backend wouldn't need to know. (Depending on whether we want other front end tools to be able to perform the SSO "registration" through the backend APIs directly.)

However if the idea is that users could continue to log in via SSO rather than creating a local server Pbench password, the backend would need to understand the SSO protocols and certainly would need to know which to use for each user.

src/pages/Profile/index.js
anishaswain: `dataExpireOptions` was a bug. Fixed it. Added `companyID` to component state

src/pages/Profile/index.js
anishaswain: added `companyID` to state

src/pages/Profile/index.js
```diff
@@ -23,90 +37,336 @@ import avatar from '../../assets/avatar.svg';
...
-  };
+      this.setState({
+        editView: !editView,
+        displayName: editedName,
+        emailVal: editedEmail,
+        positionVal: editedPosition,
+      });
+    };
+
+    this.cancleEdit = () => {
```
gurbirkalsi: Spelling error here: `this.cancelEdit`

src/pages/Profile/index.js
```diff
@@ -23,90 +37,336 @@ import avatar from '../../assets/avatar.svg';
...
+      const { editView } = this.state;
+      this.setState({
+        editView: !editView,
+      });
+    };
+
+    this.removeTag = tag => {
```
gurbirkalsi: Can we consider removing `this.removeTag` as it is not being referenced anywhere else?

src/pages/Profile/index.js
```diff
@@ -23,90 +37,336 @@ import avatar from '../../assets/avatar.svg';
 class Profile extends React.Component {
   constructor(props) {
```
gurbirkalsi, line 38: In general, methods defined in the `React.Component` constructor need to be manually binded with `this` prepended to each method name. What if we moved these methods outside of the constructor to avoid manual binding? We follow a similar design within other page components.

gurbirkalsi (changes requested):

src/pages/Profile/index.js
anishaswain: fixed!

src/pages/Profile/index.js
anishaswain: removed

gurbirkalsi (approved):
Great work @anishaswain!

PR #78, opened 2020-08-06
system comparisons page

### LINKED ISSUE #45: pbench system comparisons
Enhance the current run comparison workflow to provide system fact comparisons between workloads. Currently, `cloud.redhat.com` provides this feature via an app called Drift. This app provides comparison of  system facts like arch, bios_release_date, bios_vendor, bios_version, cloud_provider, cores_per_socket, cpu_flags,
enabled_services, fqdn, infrastructure_type, infrastructure_vendor, installed_packages, installed_services, kernel_modules, last_boot_time,
network_interfaces, number_of_cpus,  number_of_sockets, os_kernel_version, os_release, satellite_managed, system_memory, yum_repos.

For doing comparisons of configuration data relevant for performance, we'll want to extract a bit more configuration data from the sosreports (e.g. cpu type, tuned configuration, etc. etc.). The list of configuration information currently compared is at https://github.com/beav/kerlescan/blob/master/kerlescan/constants.py. The frontend for Drift can be found at https://github.com/RedHatInsights/drift-frontend. 

While enhancements are made to sosreport collection, we can implement a mocked component using PatternFly to visualize a system fact comparison.

aquibbaig:
System comparisons work.

### DIFF
--- a/config/router.config.js
+++ b/config/router.config.js
@@ -80,6 +80,11 @@ module.exports = [
         name: 'share',
         component: './SessionPlaceholder',
       },
+      {
+        path: '/system-comparisions',
+        name: 'system-comparisons',
+        component: './SystemComparisions',
+      },
     ],
   },
 ];
--- a/src/common/menu.js
+++ b/src/common/menu.js
@@ -38,6 +38,11 @@ const menuData = [
     icon: 'global',
     path: '/explore',
   },
+  {
+    name: 'System Comparisions',
+    icon: 'control',
+    path: '/system-comparisions',
+  },
 ];
 
 function formatter(data, parentPath = '', parentAuthority) {
--- a/src/pages/SystemComparisions/index.js
+++ b/src/pages/SystemComparisions/index.js
@@ -1,0 +1,190 @@
+import React, { useEffect, Fragment, useState } from 'react';
+import { Tree } from 'antd';
+import _ from 'lodash';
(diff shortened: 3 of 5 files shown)


aquibbaig:
# Demo till now:
Screenshot 2020-08-19 at 7 22 59 PM

@gurbirkalsi @portante @Anisha1234 You can notice that by default all the fields are selected on the start. The fix which I am working on right now is how to pass selected fields into the TreeView component by ant-d which would automatically mark them selected on start. That way everything would be perfect. 🚀 

Also need an option to highlight the differences in terms. Haven't found a suitable way to do that yet.

aquibbaig:
After every key has been selected by default on start, this is how it would look like:

Screenshot 2020-08-20 at 6 02 46 PM

mnmehta:
Adding a comment here based on the discussion on today's call, note that we would like to show CVE mitigation data in the systems comparison page.

config/router.config.js
```diff
@@ -80,6 +80,11 @@ module.exports = [
         name: 'share',
         component: './SessionPlaceholder',
       },
+      {
+        path: '/system-comparisions',
```
mnmehta, line 84: Typo, comparisions -> comparisons, needs fixing in multiple places

aquibbaig: @mnmehta My bad, will fix that

PR #79, opened 2020-08-12
Update Apollo Server register mutation with firstName and lastName fields

gurbirkalsi:
Apollo server registration mutation method requires firstName and lastName fields from user to successfully create User model.

The firstName and lastName fields are declared as part of the User data model in typeDefs.js and referenced within the mutation in resolvers.js.


aquibbaig (approved):
LGTM!

PR #80, opened 2020-08-17
human readable timestamps of last modified

### LINKED ISSUE #73: Formatted values of "last modified"
We store `last-modified` strings as `2020-05-20T10:38:53.334Z` which is a basic date time format we fetch from elasticsearch. What if we modified it as per user specification, for example: `5 months ago` or `20 June 2020`, a better visual cue to the users.

aquibbaig:
Making timestamps human readable across the dashboard. I will move forward with further implementations in other pages if this PR actually makes sense to the team as well. Fixes #73.

Note:
Tooltips provided by `patternfly` are incompatible with `antd table`. We need to keep this in mind when we migrate completely to the `patternfly` table implementation, we have to use the tooltip by patternfly as well. 

## Demo

Screenshot 2020-08-17 at 4 03 34 PM

### DIFF
--- a/src/e2e/controllers.e2e.js
+++ b/src/e2e/controllers.e2e.js
@@ -245,8 +245,9 @@ describe('controller page component', () => {
       '.ant-calendar-panel > .ant-calendar-footer > .ant-calendar-footer-btn > .ant-calendar-footer-extra > .ant-tag:nth-child(3)'
     );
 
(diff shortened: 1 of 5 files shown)


gurbirkalsi (approved):
Nice work @aquibbaig!

anishaswain (approved):

Issue #81, opened 2020-08-17
Inconsistent routing in the dashboard?


aquibbaig:
Moving over to either of `localhost:8000/#/` or `localhost:8000/#/dashboard` gives a 404.

### Expected behaviour

According to our umi routing configuration, it should automatically redirect to `http://localhost:8000/#/dashboard/dashboard` which seems to be a valid route. 

### Demo

<img width="1437" alt="Screenshot 2020-08-17 at 4 09 49 PM" src="https://user-images.githubusercontent.com/26324376/90387840-8642bb00-e0a4-11ea-9eae-c0b4ec0fed80.png">

PR #82, opened 2020-08-18
Integrate registerUser user authentication endpoint with SignupHandler

gurbirkalsi:
Integrate registerUser user authentication endpoint with Patternfly React registration page component

Modifies auth Redux namespace with registerUser effect and implements error handling for user registration edge cases

### DIFF
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,35 +1,40 @@
+import registerUser from '../services/auth';
+
 export default {
   namespace: 'auth',
 
   state: {
-    auth: {},
+    username: '',
   },
 
   effects: {
-    *loadUser({ payload }, { put }) {
+    *registerUser({ payload }, { call }) {
+      const response = yield call(registerUser, payload);
+      return response;
+    },
+    *loginUser({ put }) {
       yield put({
-        type: 'modifyUser',
-        payload,
+        type: 'updateUserMetadata',
       });
     },
     *logoutUser({ put }) {
       yield put({
-        type: 'removeUser',
+        type: 'removeUserMetadata',
       });
     },
   },
 
   reducers: {
-    modifyUser(state, { payload }) {
+    updateUserMetadata(state, { payload }) {
       return {
         ...state,
-        auth: payload,
+        username: payload,
       };
     },
-    removeUser(state) {
+    removeUserMetadata(state) {
       return {
         ...state,
-        auth: {},
+        username: '',
       };
     },
   },
--- a/src/pages/SignupHandler/index.js
+++ b/src/pages/SignupHandler/index.js
@@ -15,236 +15,269 @@ import { routerRedux } from 'dva/router';
 import { connect } from 'dva';
 import styles from './index.less';
 
-@connect(auth => ({
-  auth: auth.auth,
+@connect(({ datastore }) => ({
+  datastoreConfig: datastore.datastoreConfig,
 }))
 class SignupHandler extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      firstName: '',
+      lastName: '',
       username: '',
       password: '',
-      confPassword: '',
-      email: '',
-      variantVal: 'tertiary',
-      btnColor: 'black',
-      btnDisabled: 'true',
-      charLength: ['block', 'none'],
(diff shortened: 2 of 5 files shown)


aquibbaig:
@gurbirkalsi I think we need to consider #81 as a part of this PR. Routing fails because we have a path '/' for `PrivateRoute` when user is not logged in. When is user is logged in, we render the `children` in the same path which is '/' for which no component exists, since PrivateRoute renders when there is no user. What we can do is two things:

aquibbaig:
1. 
```diff
diff --git a/config/router.config.js b/config/router.config.js
index cfacb567..714ff43f 100644
--- a/config/router.config.js
+++ b/config/router.config.js
@@ -20,7 +20,7 @@ module.exports = [
     Routes: ['./src/pages/PrivateRoute/index.js'],
     routes: [
       {
-        path: '/dashboard',
+        path: '/',
         name: 'dashboard',
         icon: 'dashboard',
         component: './Controllers',
```
and then redirecting to the base path '/' to render `./Controllers` component.

2. 
Secondly, we can use some logic as mentioned in umijs [documentation](https://umijs.org/docs/routing#wrappers), i.e. redirecting to '/login' and '/signup' on auth and '/dashboard' when the user is logged in.

src/models/auth.js
```diff
@@ -1,35 +1,40 @@
...
-    *loadUser({ payload }, { put }) {
+    *registerUser({ payload }, { call }) {
+      const response = yield call(registerUser, payload);
+      return response;
+    },
+    *loginUser({ put }) {
       yield put({
-        type: 'modifyUser',
-        payload,
+        type: 'updateUserMetadata',
```
dbutenhof, line 17: Curious why you added the "metadata" ... from my perspective, I found that a bit misleading, because when I went to look at the reducer I was expecting "user-related metadata" rather than the standard "regenerate state with the new user payload". Ah, well; minor confusion...

src/pages/SignupHandler/index.js
```diff
@@ -15,236 +15,269 @@ import { routerRedux } from 'dva/router';
...
+      },
+      passwordValidationMessage: '',
+      usernameValidationMessage: '',
     };
   }
 
-  componentDidMount = () => {
-    this.disableSubmitBtn();
-  };
+  componentDidMount() {
```
dbutenhof, line 44: OK, I've used Javascript but I'm hardly an expert. I assume that converting this from a lambda binding into a function must be more than just stylistic, especially considering that the real implementation is now just another lambda binding. Why?

src/pages/SignupHandler/index.js
```diff
@@ -15,236 +15,269 @@ import { routerRedux } from 'dva/router';
...
+      submitButtonColor,
+      passwordValidation,
+      passwordValidationMessage,
+      usernameValidationMessage,
     } = this.state;
 
     const form = (
       <Form className={styles.section}>
-        <FormGroup label="First name" isRequired fieldId="horizontal-form-first-name">
+        <FormGroup label="First Name" fieldId="firstName">
```
dbutenhof, line 185: Should we really be capitalizing both words vs a normal phrase "First name"?

src/pages/SignupHandler/index.js
```diff
@@ -15,236 +15,269 @@ import { routerRedux } from 'dva/router';
...
-    if (username !== '' && email !== '' && password !== '' && confPassword !== '') {
-      this.enableSubmitBtn();
+  registerUser = () => {
+    const { dispatch, datastoreConfig } = this.props;
+    const { firstName, lastName, username, password, confirmedPassword } = this.state;
+
+    let validRegistration = true;
+    const validEmailRegex = /\S+@\S+\.\S+/;
+
+    if (!validEmailRegex.test(username)) {
```
dbutenhof, line 63: Not a big deal in the context of this change, but isn't there a better email validation package available than just the character set check? (Even more irrelevant, I suppose, since Nikhil is using a strong email validator package for the server API.)

dbutenhof (commented):
I have naive questions; but functionally this looks fine.

gurbirkalsi:
Closing in favor of #126

PR #85, opened 2020-08-31
use react memo in place of memoize-one

### LINKED ISSUE #13: Use React memo instead of memoize-one
React 16.6.0 comes up with tonnes of new features. One of them is [React Memo](https://reactjs.org/blog/2018/10/23/react-v-16-6.html). Since we use memoizeOne already we can use React Memo to decrease bundle size and also improve performance.

aquibbaig:
Fixes #13.

Notice the following from React's documentation:
```
React.memo only checks for prop changes. 
If your function component wrapped in React.memo has a useState 
or useContext Hook in its implementation, it will still rerender when state or context change.
```

So, what I think is when we change across routes in our dashboard, components won't re-render again and again, because props would be essentially the same. @gurbirkalsi What do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "jspdf": "^1.5.3",
     "lodash": "^4.17.10",
     "lodash-decorators": "^6.0.0",
-    "memoize-one": "^5.0.0",
     "moment": "^2.24.0",
     "numeral": "^2.0.6",
     "omit.js": "^1.0.0",
--- a/src/layouts/BasicLayout.js
+++ b/src/layouts/BasicLayout.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { memo } from 'react';
 import PropTypes from 'prop-types';
 import { Layout, Icon, Spin, message } from 'antd';
 import DocumentTitle from 'react-document-title';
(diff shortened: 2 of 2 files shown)


gurbirkalsi:
In general, component re-renders only occur throughout the dashboard when props have changed. `React.memo` attempts to solve the problem of memoizing the instance of the component (rather than component generation) by rendering the same component if there is an attempt to render when props have not changed. Is there an instance where this is happening in the dashboard? If so, `React.memo` would be useful for solving this issue. 

Since `React.memo` only memoizes instances of components instead of the results of function calls & arguments, we should continue to utilize `memoize-one` in order to memoize `getBreadcrumbNameMap()` since it is a pure function.

gurbirkalsi:
For more on strategies we've used to reduce component re-renders: https://github.com/distributed-system-analysis/pbench-dashboard/issues/24

aquibbaig:
> In general, component re-renders only occur throughout the dashboard when props have changed. `React.memo` attempts to solve the problem of memoizing the instance of the component (rather than component generation) by rendering the same component if there is an attempt to render when props have not changed. Is there an instance where this is happening in the dashboard? If so, `React.memo` would be useful for solving this issue.
> 
> Since `React.memo` only memoizes instances of components instead of the results of function calls & arguments, we should continue to utilize `memoize-one` in order to memoize `getBreadcrumbNameMap()` since it is a pure function.

+1 understood, closing cause invalid!

PR #86, opened 2020-09-01
Implementation of Public data view for an anonymous user

### LINKED ISSUE #59: Mock the flow for providing a Public data view for an anonymous user
We need to mock up a public data page which will be visible to the user even without logging in and will give the user visual feedback to login in order to access customised data.

Screenshot (1)

The protoype: https://marvelapp.com/dh1cd91/screen/69487661

anishaswain:
Fixes #59 
Public data page will be visible to the user even without logging in and will give the user visual feedback to login in order to access customised data.

**Mock:**
Screenshot (1)

The protoype: https://marvelapp.com/dh1cd91/screen/69487661

**Implementation**
public
public


config/router.config.js
```diff
@@ -14,71 +19,82 @@ module.exports = [
...
       },
       {
-        path: '/share/:id',
-        name: 'share',
-        component: './SessionPlaceholder',
+        path: '/private',
+        Routes: ['./src/pages/PrivateRoute/index.js'],
+        routes: [
+          {
+            path: '/private',
```
aquibbaig, line 87: Will it be okay to add a new route before each of our routes i.e. `private` followed by the regular routes. When users are authenticated, we can append their username to the route as a wildcard or something, for example:
the route when I am authenticated would become `http://<host>/{aquib}/controllers`. Idk if that would be better but we can do something similar.

src/components/AuthLayout/index.js
```diff
@@ -133,7 +231,7 @@ class AuthLayout extends Component {
             className={styles.form}
           >
             {Heading}
-            {toPreview}
+            {toPreview === undefined ? authPreview : toPreview}
```
aquibbaig: can be modified as `toPreview ? authPreview : toPreview`

aquibbaig (changes requested):
Looks great, Anisha! Good work. Just a small change to go

src/components/AuthLayout/index.js
anishaswain: it should be `toPreview ? toPreview : authPreview`  i.e `toPreview || authPreview`
Fixed!

src/pages/Overview/index.js
```diff
@@ -0,0 +1,24 @@
...
+// import { routerRedux } from 'dva/router';
+
+class Overview extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {};
+  }
+
+  componentDidMount() {}
```
gurbirkalsi: Can we consider omitting these empty lifecycle methods here?

gurbirkalsi (changes requested):
Great work @anishaswain, just one small comment!

src/pages/Overview/index.js
anishaswain: removed!

gurbirkalsi (approved):

PR #87, opened 2020-09-04
Patternfly Migration

gurbirkalsi:
Integrate PatternFly React components with Pbench dashboard

Removes dependency on ant design component library (except for Table
component) and updates unit & e2e tests with new selectors


src/assets/pbench_logo.svg
```diff
@@ -0,0 +1 @@
....56-4.33l.19-.52c.54-1.49,1.08-3,1.63-4.51l.16-.46c.57-1.56,1.13-3.14,1.71-4.73l.11-.31.78-2.17.12-.33c.58-1.61,1.16-3.24,1.74-4.87l.13-.36c.28-.76.55-1.53.83-2.3.07-.21.15-.41.22-.62.24-.69.49-1.37.74-2.06l.09-.26,1.74-4.87c.09-.24.17-.48.26-.71l.75-2.11.07-.19v-.05h0c6.33-17.74,12.7-35.71,18.09-51A5.87,5.87,0,0,0,326.59,44.8Z"/><path class="cls-1" d="M423.78,60.22a247.24,247.24,0,0,0-320-2.92l67.28,67.28a151.63,151.63,0,0,1,91-30c3.41,0,6.81.12,10.18.34C282.38,80.19,295,62,310.61,39.51a19.26,19.26,0,0,1,16-8.45,19.88,19.88,0,0,1,16.22,8.39,19.28,19.28,0,0,1,2.38,17.78C339.28,74,333,91.83,326.87,108.89a152.24,152.24,0,0,1,29.79,18.45Z"/><path class="cls-1" d="M24.5,178.47H46.07a44.76,44.76,0,0,1,20,4.56,48.72,48.72,0,0,1,12.45,8.78v-1a15.36,15.36,0,0,1,15.36-15.36h13.44A15.37,15.37,0,0,1,122,186.22a152.16,152.16,0,0,1,29.67-44.47L84.81,74.89A247.1,247.1,0,0,0,24.5,178.47Z"/><path class="cls-1" d="M442.44,78.14l-66.92,66.91A151.36,151.36,0,0,1,409.18,206a52.49,52.49,0,0,1,24,7.36,51.31,51.31,0,0,1,5.59,3.94v-26.5a15.36,15.36,0,0,1,15.36-15.36h13.44a15.36,15.36,0,0,1,15.36,15.36v17.37a49.58,49.58,0,0,1,15.19-2.33,47,47,0,0,1,7.91.65A246.53,246.53,0,0,0,442.44,78.14Z"/></g></g></svg>
```
aquibbaig: NIT: Just a small change, adding a trailing line here would make the `diff` look better.

src/pages/Search/index.js
```diff
@@ -263,115 +268,109 @@ class SearchList extends Component {
...
-                <Row style={{ display: 'flex', flexWrap: 'wrap' }}>
-                  <div style={{ marginRight: 16 }}>
-                    <p style={{ marginBottom: 0, fontSize: 12, fontWeight: 600 }}>months</p>
+                <Flex>
+                  <FlexItem style={{ marginRight: 16, marginBottom: 16 }}>
+                    <p style={{ marginBottom: 6, fontSize: 12, fontWeight: 600 }}>months</p>
                     <AntdDatePicker />
-                  </div>
+                  </FlexItem>
                   {Object.keys(mapping).map(field => (
```
aquibbaig, line 276: what happens when `mappings` are empty in the first render. In my case, when I run it locally the `Search` page breaks with an error message: 
```
Uncaught TypeError: react__WEBPACK_IMPORTED_MODULE_1__.Children.toArray(...)[0] is undefined
``` 
which is defined in line 282 `return item.split('.')[0] === field`. Have to do a null check or something like that.

src/components/Button/index.js
```diff
@@ -1,13 +1,13 @@
...
-import { Button as AntdButton } from 'antd';
+import { Button as PatternFlyButton } from '@patternfly/react-core';
 
 const Button = props => {
-  const { name, type, disabled, onClick } = props;
+  const { name, variant, disabled, onClick } = props;
 
   return (
-    <AntdButton type={type} disabled={disabled} onClick={onClick} {...props}>
+    <PatternFlyButton variant={variant} isDisabled={disabled} onClick={onClick} {...props}>
```
aquibbaig, line 8: In the file `@/components/RowSelection` and in all other place where we used Button, you would make the following changes, else we get `Invalid prop type of "primary" provided to Button. Expected "submit", "reset", "button"`.

```diff
<Button
- type="primary"
+ type="submit/reset/button"
 onClick={onCompare}
 name={compareActionName}
 disabled={!(selectedItemsLength > 0)}
 style={{ marginRight: 8 }}
  />
```

src/components/GlobalHeader/index.js
```diff
@@ -1,60 +1,44 @@
-import { Component } from 'react';
-import { routerRedux, Link } from 'dva/router';
-import { Icon, Divider, Tooltip, Alert, Badge, Tag } from 'antd';
-import { Dropdown, DropdownItem, DropdownToggle, Avatar } from '@patternfly/react-core';
-import Debounce from 'lodash-decorators/debounce';
+import React, { Component } from 'react';
+import { routerRedux } from 'dva/router';
+import {
+  PageHeader,
+  PageHeaderTools,
```
aquibbaig, line 5: Is some error as follows thrown when importing `PageHeaderTools` in your system? I am pretty sure this warning is thrown during import of one of these: PageHeaderTools, PageHeaderToolsGroup.
```
Warning: React.createElement: type is invalid -- expected a string 
(for built-in components) or a class/function (for composite components)
 but got: undefined. You likely forgot to export your component from
 the file it's defined in, or you might have mixed up default and named imports.
```

aquibbaig (changes requested):
did an initial review on this Pull request. The design looks great @gurbirkalsi. This is a priority one task. I've added a few comments, just small changes. Apart from that, I think the `GlobalHeader` does not work in the Search Page.

src/pages/Search/index.js
gurbirkalsi: Unfortunately, I wasn't able to reproduce this issue locally. `mapping` should default to an empty object specified by its default Redux store value. `selectedFields` should populate as an array of arrays containing config data. Would you be able `console.log` the state of those values where this error occurs and the full stack trace?

src/components/Button/index.js
gurbirkalsi: Which version of Patternfly React is installed to your node modules? The latest documentation notes ```'primary' | 'secondary' | 'tertiary' | 'danger' | 'link' | 'plain' | 'control'``` as potential button variants.

src/components/GlobalHeader/index.js
gurbirkalsi: I wasn't able to reproduce the issue on my local build; however, I would try upgrading your Patternfly React dependency as it might be referencing a previous version of the library.

aquibbaig: @gurbirkalsi My dependencies are at
```
"@patternfly/react-charts": "^5.3.19",
"@patternfly/react-core": "^3.153.2",
"@patternfly/react-icons": "^3.15.16",
"@patternfly/react-table": "^2.28.39",
```

src/components/Button/index.js
aquibbaig: I think that may be because I might as well be using an older patternfly-react version

src/pages/Search/index.js
aquibbaig: so, i logged `mappings` and  the above error appeared when mappings were empty, so it couldn't find the zeroth element. The stack trace is described in the above comment.

aquibbaig (approved):
LGTM! :tada:

PR #75, opened 2020-07-24
Dashboard APIs

FuqingWang:
Replaced all Elasticsearch calls to use pbench server APIs
Server address {pbench_server) needs to be specified in datastoreConfig.js to enable API calls
References for Elasticsearch and GraphQL from datastoreConfig.js have been moved to the server(pbench-server.cfg).

PR #95, opened 2020-10-23
Dashboard to work with Server APIs 

npalaska:
This is a followup improvement PR on top of https://github.com/distributed-system-analysis/pbench-dashboard/pull/75

> Replaced all Elasticsearch calls to use pbench server APIs
Server address {pbench_server) needs to be specified in datastoreConfig.js to enable API calls
References for Elasticsearch and GraphQL from datastoreConfig.js have been moved to the server(pbench-server.cfg)

### DIFF
--- a/src/services/dashboard.js
+++ b/src/services/dashboard.js
@@ -7,11 +7,15 @@ import { getAllMonthsWithinRange } from '../utils/moment_constants';
 const { endpoints } = window;
 
 function scrollUntilEmpty(data) {
-  const endpoint = `${endpoints.elasticsearch}/_search/scroll?scroll=1m`;
+  const url = `/_search/scroll?scroll=1m`;
+  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
   const allData = data;
+  const indices = `_search/scroll?scroll=1m&scroll_id=${allData._scroll_id}`;
 
   if (allData.hits.total.value !== allData.hits.hits.length && allData._scroll_id) {
-    const scroll = request.post(`${endpoint}&scroll_id=${allData._scroll_id}`);
+    const scroll = request.post(endpoint, {
+      data: { indices },
+    });
     scroll.then(response => {
       allData._scroll_id = response._scroll_id;
       allData.hits.total = response.hits.total;
@@ -45,41 +49,46 @@ export async function queryResults(params) {
   try {
     const { selectedDateRange, controller } = params;
 
-    const endpoint = `${endpoints.elasticsearch}/${getAllMonthsWithinRange(
+    const indices = `${getAllMonthsWithinRange(
       endpoints,
       endpoints.run_index,
       selectedDateRange
     )}/_search`;
 
+    const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+
     return request.post(endpoint, {
-      params: {
-        ignore_unavailable: true,
-      },
       data: {
-        _source: {
-          includes: [
-            '@metadata.controller_dir',
-            '@metadata.satellite',
-            'run.controller',
-            'run.start',
-            'run.end',
-            'run.name',
-            'run.config',
-            'run.prefix',
-            'run.id',
-          ],
+        indices,
+        params: {
+          ignore_unavailable: true,
         },
-        sort: {
-          'run.end': {
-            order: 'desc',
+        payload: {
+          _source: {
+            includes: [
+                '@metadata.controller_dir',
+                '@metadata.satellite',
+                'run.controller',
+                'run.start',
+                'run.end',
+                'run.name',
+                'run.config',
+                'run.prefix',
+                'run.id',
+            ],
           },
-        },
-        query: {
-          match: {
-            'run.controller': controller[0],
+          sort: {
+            'run.end': {
+              order: 'desc',
+            },
           },
+          query: {
+            match: {
+              'run.controller': controller[0],
+            },
+          },
+          size: 5000,
         },
-        size: 5000,
       },
     });
   } catch (error) {
@@ -90,23 +99,28 @@ export async function queryResults(params) {
 export async function queryResult(params) {
   const { selectedDateRange, result } = params;
 
-  const endpoint = `${endpoints.elasticsearch}/${getAllMonthsWithinRange(
+  const indices = `${getAllMonthsWithinRange(
     endpoints,
     endpoints.run_index,
     selectedDateRange
   )}/_search`;
 
+  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+
   return request.post(endpoint, {
-    params: {
-      ignore_unavailable: true,
-    },
     data: {
-      query: {
-        match: {
-          'run.name': result,
-        },
+      indices,
+      params: {
+        ignore_unavailable: true,
       },
-      sort: '_index',
+      data: {
+        query: {
+          match: {
+            'run.name': result,
+          },
+        },
+        sort: '_index',
+      },
     },
   });
 }
@@ -114,15 +128,20 @@ export async function queryResult(params) {
 export async function queryTocResult(params) {
   const { selectedDateRange, id } = params;
 
-  const endpoint = `${endpoints.elasticsearch}/${getAllMonthsWithinRange(
+  const indices = `${getAllMonthsWithinRange(
     endpoints,
     endpoints.run_toc_index,
     selectedDateRange
   )}/_search?q=run_data_parent:"${id}"`;
 
+  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+
   return request.post(endpoint, {
-    params: {
-      ignore_unavailable: true,
+    data: {
+      indices,
+      params: {
+        ignore_unavailable: true,
+      },
     },
   });
 }
@@ -130,45 +149,50 @@ export async function queryTocResult(params) {
 export async function queryIterationSamples(params) {
   const { selectedDateRange, selectedResults } = params;
 
-  const endpoint = `${endpoints.elasticsearch}/${getAllMonthsWithinRange(
+  const indices = `${getAllMonthsWithinRange(
     endpoints,
     endpoints.result_index,
     selectedDateRange
   )}/_search?scroll=1m`;
 
+  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+
   const iterationSampleRequests = [];
   selectedResults.forEach(run => {
     iterationSampleRequests.push(
       request.post(endpoint, {
-        params: {
-          ignore_unavailable: true,
-        },
         data: {
-          size: 1000,
-          query: {
-            match: {
-              'run.id': run.id,
-            },
+          indices,
+          params: {
+            ignore_unavailable: true,
           },
-          aggs: {
-            id: {
-              terms: {
-                field: 'run.id',
+          data: {
+            size: 1000,
+            query: {
+              match: {
+                'run.id': run.id,
               },
-              aggs: {
-                type: {
-                  terms: {
-                    field: 'sample.measurement_type',
-                  },
-                  aggs: {
-                    title: {
-                      terms: {
-                        field: 'sample.measurement_title.raw',
-                      },
-                      aggs: {
-                        uid: {
-                          terms: {
-                            field: 'sample.uid',
+            },
+            aggs: {
+              id: {
+                terms: {
+                  field: 'run.id',
+                },
+                aggs: {
+                  type: {
+                    terms: {
+                      field: 'sample.measurement_type',
+                    },
+                    aggs: {
+                      title: {
+                        terms: {
+                          field: 'sample.measurement_title',
+                        },
+                        aggs: {
+                          uid: {
+                            terms: {
+                              field: 'sample.uid',
+                            },
                           },
                         },
                       },
@@ -176,26 +200,26 @@ export async function queryIterationSamples(params) {
                   },
                 },
               },
-            },
-            name: {
-              terms: {
-                field: 'run.name',
+              name: {
+                terms: {
+                  field: 'run.name',
+                },
+              },
+              controller: {
+                terms: {
+                  field: 'run.controller',
+                },
               },
             },
-            controller: {
-              terms: {
-                field: 'run.controller',
+            sort: [
+              {
+                'iteration.number': {
+                  order: 'asc',
+                  unmapped_type: 'boolean',
+                },
               },
-            },
+            ],
           },
-          sort: [
-            {
-              'iteration.number': {
-                order: 'asc',
-                unmapped_type: 'boolean',
-              },
-            },
-          ],
         },
       })
     );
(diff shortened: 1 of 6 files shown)


npalaska:
How do we update the `datastoreConfig` file for the travis CI runs?

aquibbaig:
> How do we update the `datastoreConfig` file for the travis CI runs?

Hi @npalaska. I dont think you can update or you need to do that. The endpoints contained in datastoreConfig are actually replaced by mock files in the test suite. Also, the tests are failing, let's debug them together

npalaska:
@aquibbaig are you talking about the `e2etests` file where we export the `datastoreConfig` values to `mock/datastoreConfig.js`,? but in that case we would have to expose our internal `pbench_server` endpoint isnt it

aquibbaig:
> @aquibbaig are you talking about the `e2etests` file where we export the `datastoreConfig` values to `mock/datastoreConfig.js`,? but in that case we would have to expose our internal `pbench_server` endpoint isnt it

Yes @npalaska. The datastoreConfig remains private as your own local configuration. We don't commit it anywhere, in e2e tests the values need to be fetched from the mock file rather than any endpoints specified in the `datastoreConfig.js`. Considering the fact that now the urls have been modified after migrating to pbench server, you might wanna check if we are still able to intercept requests in each page and replace the data contained in those urls with those present in our mock file.

src/services/search.js
```diff
@@ -4,56 +4,61 @@ import request from '../utils/request';
 export async function queryIndexMapping(params) {
   const { datastoreConfig, indices } = params;
 
-  const endpoint = `${datastoreConfig.elasticsearch}/${datastoreConfig.prefix}${
-    datastoreConfig.run_index
-  }${indices[0]}/_mappings`;
+  const url = `${datastoreConfig.prefix}${datastoreConfig.run_index}${indices[0]}/_mappings`;
```
dbutenhof: This isn't a `url`. It's not really an "index" either since you're pulling the template mappings for one index.

I'm not quite sure _what_ to call this, actually; although in terms of the interim pass-through Elasticsearch API, we call it "indices".

dbutenhof (commented):
We need to fix up the Ansible dashboard deploy/install mechanism to correctly generate the dashboard config file from an updated `config.json.j2` template.

I'm also not that thrilled about separating this "essential cleanup" from the original PR, but that's a technicality.

src/services/search.js
portante: Perhaps we should separate this out a bit:

 * The code forms a full index name: `${datastoreConfig.prefix}${datastoreConfig.run_index}${indices[0]}`
 * The `url` is really trying to specify the `_mapping` API for a given index

portante (changes requested):

src/e2e/controllers.e2e.js
```diff
@@ -37,7 +37,7 @@ beforeAll(async () => {
         headers: { 'Access-Control-Allow-Origin': '*' },
         body: JSON.stringify(generateMockControllerAggregation),
       });
-    } else if (request.method() === 'GET' && request.url().includes('aliases')) {
+    } else if (request.method() === 'POST' && request.postData().includes('aliases')) {
```
dbutenhof: I'd prefer the JSON payload check to be a bit more specific than just having "aliases" anywhere. Though most likely the test environment won't ever see it anywhere but the expected "data"... and, OK, the adjacent tests aren't any more specific. :-(

The repetition of this pattern so many times seems awkward, too; what's really happening is trying to identify a query from the HTTP method and a "unique" URI or some snippet of JSON payload. This could probably be refactored and cleaned up (and made more robust); but despite griping about it, I'm not going to push you to attempt that now.

src/services/search.js
```diff
@@ -4,11 +4,13 @@ import request from '../utils/request';
 export async function queryIndexMapping(params) {
   const { datastoreConfig, indices } = params;
 
-  const url = `${datastoreConfig.prefix}${datastoreConfig.run_index}${indices[0]}/_mappings`;
+  const mappingIndix = `${datastoreConfig.prefix}${datastoreConfig.run_index}${
```
dbutenhof: Oops; `mappingIndex`??

src/e2e/controllers.e2e.js
npalaska: The JSON payload is just a small mock payload that we return for the purpose of e2e tests and I am not sure but I think that might have been the original reason behind this checks, I can definitely clean this up a little but this is again going to be modified when we have query specific REST APIs

src/services/search.js
npalaska: yeah need to think something better

aquibbaig: Simply, `mappings` should be fine

aquibbaig (approved):
Looks good, @npalaska! Great work! :tada:

src/services/search.js
dbutenhof: I was really just commenting on the typo. I wouldn't object to a more clear name, if anyone can think of one.

dbutenhof (approved):

src/e2e/controllers.e2e.js
portante: What is the reason for the change from `GET` to `POST`?

portante (approved):
Just wondering about the `GET` to `POST` reason ...

npalaska:
@portante Its in accordance with the changes we have in our Server API, if you look at the `/elasticsearch` endpoint in server API there is no `GET` method only `POST` request. We decide whether to do the get request based on the condition of payload in the data, if payload is present in the data we perform the post otherwise if payload is missing then we just perform the get request and return the response we got from the elasticsearch 
To which I think should have had a proper `GET` method in our API but we are anyway moving away from the API calls to direct `elasticsearch` endpoint to support more targeted queries

src/services/dashboard.js
```diff
@@ -38,41 +41,46 @@ export async function queryControllers(params) {
...
-      },
       data: {
-        filter: {
-          range: {
-            '@timestamp': {
-              gte: selectedDateRange.start,
-              lte: selectedDateRange.end,
-            },
-          },
+        indices,
```
dbutenhof: This is more a comment on the original Pbench server API design, but it didn't really start bugging me until I saw it appear repeatedly in the dashboard changes.

This is not well-formed JSON, because we're just throwing in an unlabeled string. This ought to be another key, like "uri" ... although `uri` is misleading since it's really only the resource part of the uri. So I suppose `"resource"` or `"target_resource"` would be better.

I'm not inclined to push this issue too far, since these changes are short lived, and when we have a real API we'll be cutting off the pass-through APIs anyway. Still, it's going to continue to bug me every time I see it.

dbutenhof (approved):

src/services/dashboard.js
npalaska: I was trying to make eslint happy with https://eslint.org/docs/rules/object-shorthand. Apparently in ES6 `var foo = {x, y, z};` = `var foo = {x:x, y:y, z:z};`

dbutenhof: Ah; and I knew that, too, but I clearly hadn't completely twisted my brain from Python to Javascript! Sorry 'bout that.

On the other hand, a JSON key of "indices" isn't really accurate, either, but that's not a big deal in this context.

gurbirkalsi (approved):

anishaswain (approved):

dbutenhof:
This really needs to be rebased and brought up to date!

src/services/dashboard.js
```diff
@@ -7,11 +7,15 @@ import { getAllMonthsWithinRange } from '../utils/moment_constants';
 const { endpoints } = window;
 
 function scrollUntilEmpty(data) {
-  const endpoint = `${endpoints.elasticsearch}/_search/scroll?scroll=1m`;
+  const url = `/_search/scroll?scroll=1m`;
```
dbutenhof: I don't think `url` is used, and it's redundant with `indices` below.

src/services/search.js
```diff
@@ -6,55 +6,61 @@ const { endpoints } = window;
...
 
-  const endpoint = `${endpoints.elasticsearch}/${endpoints.prefix}${endpoints.run_index}${
-    indices[0]
-  }/_mappings`;
-  return request.get(endpoint);
+  const mappings = `${endpoints.prefix}${endpoints.run_index}${indices[0]}/_mappings`;
+
+  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+
+  return request.post(endpoint, { data: { indices: mappings } });
```
dbutenhof, line 13: Why not just call it `indices` instead of `mappings`? (I don't think the key "indices" is really any more accurate anywhere else than here, and it's just a name.)

But, no big deal, and no need to change it.

src/services/dashboard.js
```diff
@@ -258,50 +287,55 @@ export async function queryIterationSamples(params) {
 export async function queryTimeseriesData(payload) {
   const { selectedDateRange, selectedIterations } = payload;
 
-  const endpoint = `${endpoints.elasticsearch}/${getAllMonthsWithinRange(
+  const indices = `${parseMonths(
```
dbutenhof: Why the change from `getAllMonthsWithinRange` to `parseMonths`?

dbutenhof: > Why the change from `getAllMonthsWithinRange` to `parseMonths`?

Ah; and that's your unit test failure, because there _is_ no `parseMonths`.

npalaska: This is interesting I dont seem to recall how this change came while rebasing

portante (commented):
General direction looks good.  I'll review again once @dbutenhof's comments have been addressed.

aquibbaig:
Hey @npalaska! Can you rebase this branch on top of `master` which will probably fix those tests for you

src/services/dashboard.js
portante: @npalaska ?

src/services/dashboard.js
portante: @npalaska ?

portante (changes requested):
Needs a rebase and outstanding comments addressed and resolved.

riya-17 (approved):
Looks Good

dbutenhof:
Can we pick this up again? It should be close to ready with a rebase and minor adjustments; it would get the dashboard away from direct Elasticsearch access. (The GraphQL direct access will go away in any case as we migrate to Pbench server authentication and metadata, but this seems like a good interim step.)

src/services/dashboard.js
```diff
@@ -90,112 +98,127 @@ export async function queryResults(params) {
...
+                },
+                aggs: {
+                  type: {
+                    terms: {
+                      field: 'sample.measurement_type',
+                    },
+                    aggs: {
+                      title: {
+                        terms: {
+                          field: 'sample.measurement_title',
```
dbutenhof: You lost the `.raw` here; `sample.measurement_title` isn't "aggregatable" (`"text"`) but `sample.measurement_title.raw` is (`"keyword"`)

dbutenhof (changes requested):
Mostly good, but one critical transcription error.

src/services/dashboard.js
npalaska: Ah good catch

dbutenhof (approved):
Reminder... this would have been easier to re-review if you'd pushed a separate commit to fix the problem, which would allow reviewers to be more confident they were seeing all changes without re-reviewing _everything_! ;-)

npalaska:
> Reminder... this would have been easier to re-review if you'd pushed a separate commit to fix the problem, which would allow reviewers to be more confident they were seeing all changes without re-reviewing _everything_! ;-)

Yes; it was just 1 line change so I decided to squash it in the previous commit. I also didnt want to get approval and then discard all the approvals because of the rebase, but interestingly it didnt happen here with my previous rebase, huh !!

src/services/dashboard.js
```diff
@@ -7,11 +7,14 @@ import { getAllMonthsWithinRange } from '../utils/moment_constants';
 const { endpoints } = window;
 
 function scrollUntilEmpty(data) {
-  const endpoint = `${endpoints.elasticsearch}/_search/scroll?scroll=1m`;
+  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
   const allData = data;
+  const indices = `_search/scroll?scroll=1m&scroll_id=${allData._scroll_id}`;
```
portante: Usually, the first `scroll` API call established a context, returning a `scroll_id`, which is used on subsequent calls.  Is that still happening?

npalaska: Not sure what you meant by that, this is just a structural change separating an es endpoint from indices without any logical changes. In the original code we are querying like this
```
const endpoint = `${endpoints.elasticsearch}/_search/scroll?scroll=1m`;
const scroll = request.post(`${endpoint}&scroll_id=${allData._scroll_id}`);
```
However with server api we need a `data` field with a post payload so we do the following specifying the indices separately 
```
endpoint = `${endpoints.pbench_server}/elasticsearch`;
const indices = `_search/scroll?scroll=1m&scroll_id=${allData._scroll_id}`;
const scroll = request.post(endpoint, {
      data: { indices },
    });

Comment (portante) [reply src/services/dashboard.js]: Isn't there an `if` on line 13 where the `scroll_id` is only added if it exists in the returned data?

The `scroll_id` seems to only be added at line 19 below.

**Review by portante — CHANGES_REQUESTED** [review]
Seems like we need to correct the use of the `scroll_id` still, but looks good otherwise.

**Comment by dbutenhof** [comment]
> Seems like we need to correct the use of the `scroll_id` still, but looks good otherwise.

Need to figure out whether this is actually the case and if so, get in a fix. I'd like to get this merged finally, if we can!

File: src/services/dashboard.js (hunk 9 above)
Comment (dbutenhof) [reply src/services/dashboard.js]: I think that Javascript will just return `undefined` at line 12 if the `_scroll_id` property is missing, and in that case the `if` at line 14 won't trigger, so the `&scroll_id=undefined` won't be used.

But perhaps Gurbir or Aquib can confirm the Javascript behavior more confidently...

Comment (dbutenhof) [reply src/services/dashboard.js]: Actually, I just pulled up Chrome's debug console and tried it; the undefined JSON field does simply resolve as "undefined":

```
x = {a: "test", b: "foo"}
y = `?scroll_id=${x.bar}`

"?scroll_id=undefined"
```

So I'm pretty sure this is fine.

portante: What does the Elasticsearch V7 scroll API do when it receives `undefined` as the `scroll_id`?

Aside from the fact that specifying the `scroll_id` in the parameters of the URL is deprecated as of 7.0, has anybody tested this change to make sure it behaves as expected?

See https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html.

dbutenhof: But at line 14, `allData._scroll_id` will be False and we won't POST the request, so that doesn't matter.

On the other hand, sure, the code would be a tiny bit more efficient if we assembled the potential URI within the conditional.

npalaska: Okay I fixed it by creating indices inside if block so `scroll_id` should never be undefined in the `indices`

portante: Okay, it appears that `scrollUntilEmpty` is only ever called after an initial `_search/scroll` API is made, so this appears to be correct.

Though, it would be nice if somebody can confirm this actually still works.

portante (approved):
Let's hope the scroll API still works ...

src/services/dashboard.js
dbutenhof: I just did some experimenting with direct Elasticsearch queries, and id `15ed6b30e0f92df4fffdba7bae72c41d` (`fio_rw_2021.03.03T02.04.39`) on the staging server is returning a `_scroll_id` for `queryIterationSamples`, so the code ought to be triggering if Nikhil runs his dashboard code and drills down through that dataset.

gurbirkalsi:
@npalaska and I discovered an issue where the `ignore_unavailable:true` parameter for each request to Elasticsearch was not encoded correctly within the request URL to the Pbench API server. Nikhil is going to open a pull request on the server to address this encoding issue. We were able to visualize data for various workflows where data is available on staging and can merge this PR!

PR #97, opened 2020-10-28
favorite/unfavorite records for registered users

### LINKED ISSUE #94: Unable to remove favorite records in the table
Once we click on the `star` icon, we are able to favorite records to our redux store and also add them to a separate favorites panel. Once the icon is clicked, it gets disabled, i think the user would be able to remove a specific record from favorites as well.

aquibbaig:
This PR introduces the following changes
- Only registered users should be able to bookmark records and remove them.
- Fixes #94

### DIFF
--- a/src/components/LoginModal/index.js
+++ b/src/components/LoginModal/index.js
@@ -1,0 +1,29 @@
+import React, { useEffect, useState } from 'react';
+import Button from '@/components/Button';
+import { Modal, ModalVariant } from '@patternfly/react-core';
+
+export default props => {
+  const [isModalOpen, setOpenModal] = useState(false);
+  useEffect(() => {
+    const { showLoginModal } = props;
+    setOpenModal(showLoginModal);
+  });
+  const { redirect, content, onConfirm, modalTitle } = props;
+  const handleModalToggle = () => {
+    setOpenModal(false);
+    onConfirm();
+  };
+  return (
+    <React.Fragment>
+      <Modal
+        variant={ModalVariant.small}
+        isOpen={isModalOpen}
+        onClose={handleModalToggle}
+        title={modalTitle || ' '}
+        actions={[<Button key="confirm" variant="primary" onClick={redirect} name="confirm" />]}
+      >
+        {content}
+      </Modal>
+    </React.Fragment>
+  );
+};
--- a/src/pages/Controllers/index.js
+++ b/src/pages/Controllers/index.js
@@ -16,17 +16,19 @@ import {
 
 import SearchBar from '@/components/SearchBar';
 import AntdDatePicker from '@/components/DatePicker';
+import LoginModal from '@/components/LoginModal';
 import Table from '@/components/Table';
 import { getDiffDate } from '@/utils/moment_constants';
 
 const { TabPane } = Tabs;
 
-@connect(({ datastore, global, dashboard, loading, user }) => ({
+@connect(({ datastore, global, dashboard, loading, user, auth }) => ({
   controllers: dashboard.controllers,
   indices: datastore.indices,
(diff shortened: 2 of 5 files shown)


src/pages/Results/index.js
```diff
@@ -143,12 +144,20 @@ class Results extends Component {
...
+
+    if (username) {
+      const { dispatch } = this.props;
+      // dispatch an action to favorite controller
+      dispatch({
+        type: 'user/favoriteController',
+        payload: result,
+      });
+    } else {
+      console.log('Not logged on');
```
aquibbaig: Ideally, we would like to show a modal with a message that says "Register or Login into pbench dashboard" and a link redirecting to the login page. @anishaswain If there already exists one in our user auth PR, we can reuse that component?

anishaswain: @aquibbaig we do not have any modal component having the option to redirect to login page yet

aquibbaig: @anishaswain We can make one and reuse the same for the login component

src/components/LoginModal/index.js
```diff
@@ -0,0 +1,32 @@
...
+export default props => {
+  const [isModalOpen, setOpenModal] = useState(false);
+  useEffect(() => {
+    const { showLoginModal } = props;
+    setOpenModal(showLoginModal);
+  });
+  const { redirect, content, callback, modalTitle } = props;
+  const handleModalToggle = () => {
+    setOpenModal(false);
+    callback();
```
gurbirkalsi: Nice use of hooks! Can we consider a more descriptive name for our callback function to improve readability? In this case, our callback references `this.modalConfirm()`. What if we considered a naming schema with the `on` prefix (in this case `onConfirm`)?

src/components/LoginModal/index.js
```diff
@@ -0,0 +1,32 @@
+import React, { useEffect, useState } from 'react';
+import { Modal, ModalVariant, Button } from '@patternfly/react-core';
```
gurbirkalsi: We should consider referencing the `Button` component definition already defined within the dashboard (contains simple abstractions to simplify dropping the component into markup).

src/pages/Controllers/index.js
```diff
@@ -232,6 +284,12 @@ class Controllers extends Component {
                   />
                   <AntdDatePicker onChangeCallback={this.fetchControllers} />
                 </Form>
+                <LoginModal
+                  showLoginModal={showLoginModal}
+                  redirect={this.redirectToLoginPage}
+                  content="Please login on to pbench dashboard to bookmark records."
```
gurbirkalsi: Can we consider simplifying this message to `Please login to favorite controllers`?

gurbirkalsi (changes requested):

aquibbaig:
@gurbirkalsi Rather than having multiple definitions of `LoginModal` for each page, will it make sense to add a single definition with a message "You need to login in order to do that action" in our `BasicLayout` component definition which is a common parent of all our page components?

src/pages/Controllers/index.js
aquibbaig: yes

gurbirkalsi:
That approach sounds great! We should be able to pass a message along with the event handler in `showLoginRedirectModal()`.

anishaswain (commented):
@aquibbaig the PR seems to be behind master by 6 commits and hence missed the routing changes from #86 . A simple rebase should fix that.

aquibbaig:
Since the latest discussion, we do not want to show a modal component right now, this will surely be a priority once we have set up user integration in the backend, hence a workaround for now would be to just hide the "actions" column for non-registered users. Tracked in #108

PR #105, opened 2020-11-16
Addition of Overview page in the dashboard

### LINKED ISSUE #90: Design overview page for logged in user
We need to mock up a landing page for visualizing new unprocessed pbench runs (both private and public) when the user logs in. 

**Concept:**
1. The new data sent by the user could never be public
2.  When the new data comes in the user should be able to see it in their dashboard profile after logging in
3. The user has to decide the privacy of the data with the given call to action options.  
4. The user needs to have a call to action option for preserving data, deleting the data or making it public
5. The user should be aware of the data getting expired soon

Jira: PBENCH-16

### LINKED ISSUE #113: Clicking on "Dashboard" in sidebar opens up the private view
I am not sure it is the desired behaviour but clicking on the "dashboard" link in the sidebar opens up the private view. Note that when we open the application for the first time, the "dashboard" link is active, meaning we are in the Controllers page.

anishaswain:
Fixes #90 , #113 
Preview till now:
Overview-Pbench-Dashboard

- [ ] Favouriting feature should use result keys
- [ ] Result should be marked as seen upon clicking on them.


mock/overview.js
```diff
@@ -4,67 +4,67 @@ export const expirationLimit = 20;
 export const resultData = [
   {
     key: 1,
-    result: ['fio_2020.09.10T12.42.11', ' virtual506.virt.lab.eng.bos.redhat.com'],
+    result: ['mock result -1', ' virtual506.virt.lab.eng.bos.redhat.com'],
```
portante: Damage is done, but we should probably remove internal host names and replace them with "*.example.com" or something.

portante (changes requested):
Looks good, just a bit of clean up.

aquibbaig:
@portante We can edit the commit that specifically adds those host names

src/pages/PrivateRoute/index.js
```diff
@@ -1,21 +1,20 @@
 import React, { Component, Fragment } from 'react';
-import AuthLayout from '@/components/AuthLayout';
-import { routerRedux } from 'dva/router';
+// import AuthLayout from '@/components/AuthLayout';
+// import { routerRedux } from 'dva/router';
```
portante: Do we need these commented out?

src/common/menu.js
```diff
@@ -2,7 +2,8 @@ import { isUrl } from '../utils/utils';
 
 const menuData = [
   {
-    name: 'Dashboard',
+    name: 'Overview',
+    icon: 'dashboard',
```
gurbirkalsi: The `SiderMenu` component doesn't currently render an icon with the associated route. We can either add support by extending the component or remove this field for now.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+import {
+  OutlinedClockIcon,
+  UndoAltIcon,
+  ExternalLinkAltIcon,
+  EllipsisVIcon,
+} from '@patternfly/react-icons';
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import { resultData, expirationLimit } from '../../../mock/overview';
+import Table from '@/components/Table';
+// import { Table } from 'antd';
```
gurbirkalsi: We can remove this comment for importing `Table`.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+      newData: [],
+      unlabledData: [],
+    };
   }
 
+  componentDidMount() {
+    this.getSeperatedResults();
+  }
+
+  getSeperatedResults() {
```
gurbirkalsi: Once this data is available via the API, we can probably remove a lot of this post processing (as it moves to the backend).

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+  onCollapse = (event, rowKey, isOpen) => {
+    const { rows } = this.state;
+    rows[rowKey].isOpen = isOpen;
+    this.setState({
+      rows,
+    });
+  };
+
+  onFavourite = e => {
+    console.log(e);
```
gurbirkalsi: In general, we want to avoid logging in production code. We can consider removing this functionality until the necessary hooks are in place to send a request via the API.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+  };
+
+  showDrowpdown = id => {
+    const dropdownElem = document.getElementById(id);
```
gurbirkalsi, line 150: We should avoid using the DOM API where the PatternFly component API exposes this functionality in order to preserve DOM integrity. Can we consider utilizing the `Dropdown` component? (https://www.patternfly.org/v4/components/dropdown/)

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+        title: 'End Time',
+        dataIndex: 'end',
+        key: 'end',
+      },
+      {
+        title: 'Scheduled for deletion on',
+        dataIndex: 'deletion',
+        key: 'deletion',
+        render: text => {
+          const deleteDate = moment(new Date(Date.parse(text)));
```
gurbirkalsi: Is it possible to use the `getTimeDistance()` method in `utils.js` here?

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+              </div>
+            </div>
+          );
+        },
+      },
+    ];
+
+    const rowSelection = {
+      onChange: (selectedRowKeys, selectedRows) => {
+        console.log(`selectedRowKeys: ${selectedRowKeys}`, 'selectedRows: ', selectedRows);
```
gurbirkalsi: See comment regarding removing console.log() from checked in code.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+      getCheckboxProps: record => ({
+        disabled: record.name === 'Disabled User', // Column configuration not to be checked
+        name: record.name,
+      }),
+    };
+
     return (
-      <div>
-        <h1>Overview Page</h1>
+      <div className={styles.paddingBig}>
```
gurbirkalsi, line 575: Once we've collected more feedback regarding the Overview design, we should consider moving markup in subcomponents in order to improve organization and performance.

gurbirkalsi (changes requested):

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+                />
+              </div>
+            );
+          }
+          return (
+            <div>
+              <span>
+                {moment(text)
+                  .add(7, 'days')
+                  .format('YYYY-MM-DDTHH:mm:ss:SSSSSS')}
```
aquibbaig: Hey, we can use a helper function in utils.js

src/pages/Overview/index.js
```diff
@@ -1,16 +1,880 @@
...
+                                    { name: 'meta1', x: '7-10-2020', y: 70 },
+                                    { name: 'meta1', x: '8-10-2020', y: 10 },
+                                    { name: 'meta2', x: '9-10-2020', y: 50 },
+                                    { name: 'meta2', x: '10-10-2020', y: 50 },
+                                    { name: 'meta2', x: '11-10-2020', y: 30 },
+                                    { name: 'meta3', x: '12-10-2020', y: 50 },
+                                    { name: 'meta2', x: '13-10-2020', y: 50 },
+                                    { name: 'meta2', x: '14-10-2020', y: 30 },
+                                    { name: 'meta3', x: '15-10-2020', y: 150 },
+                                    { name: 'meta2', x: '16-10-2020', y: 50 },
```
aquibbaig: Can we reference this as an object defined in a mock file or it wont work that way, I'm not sure about it?

aquibbaig (changes requested):
Some minor changes, but I believe it's coming along pretty well!

src/pages/LoginHandler/index.js
```diff
@@ -1,146 +1,11 @@
...
-            </Button>
-          </ActionGroup>
-        </Form>
-      </div>
-    );
-    return <AuthLayout toPreview={form} heading="Log into your Pbench Acount" backOpt="true" />;
-  }
+function LoginHandler() {
+  return (
+    <AuthLayout toPreview={<LoginForm />} heading="Log into your Pbench Acount" backOpt="true" />
```
portante: `s/Acount/Account/`

portante (changes requested):

src/pages/Overview/index.js
anishaswain: @gurbirkalsi `getSeperatedResults()` functions filter the seen and unseen results by the user. I believe, the backend is only going to provide us with a flag referencing to the seen status but in order to  display the results in separate tables, we still need filter the object according to the seen flag and create two separate objects for both tables.

portante (commented):
Where are we at with addressing the previous comments?

src/pages/Overview/index.js
anishaswain: @gurbirkalsi the overview page still uses ant design table component. And the table version that we are using in ant design doesn't support multiple dropdown actions in table. The PatternFly dropdown component would be a perfect fit here but if I am trying to use both AntD and PF then it is resulting in a conflict. So I am waiting for #124 PR to get in in order to make changes. Or, should I make change the table component to PF by removing Ant design in this PR itself?

src/pages/Overview/index.js
anishaswain: @gurbirkalsi as per my understanding, `getTimeDistance()` function is returning two moment objects having time distance of a year, month or days. Here we are trying to calculate the number of days remaining for the result deletion and the deletion date is being imported from mock value. I am not really sure how to incorporate `getTimeDistance()` function here. However I have created a separate function in `moment_constants.js` file to get days difference between two dates. Let me know if it works.

anishaswain:
> Where are we at with addressing the previous comments?

@portante I have addressed all the comments now.

src/pages/Overview/index.js
anishaswain: Removed this. Didn't need formatting.

portante (approved):

portante (changes requested):
Actually, the TravisCI build is failing.

src/pages/Overview/index.js
gurbirkalsi: Sorry, I didn't quite understand what you meant by "multiple dropdown actions in table". PatternFly supports a Menu component which can be utilized to display menu actions for each table row. We should be able to use the component abstraction to access selected Menu items without having to reference the DOM.

src/pages/RunResults/index.js
```diff
@@ -0,0 +1,71 @@
...
+  selectedPrivateController: privatedatastore.selectedPrivateController,
+  privateControllers: privatedatastore.privateControllers,
+}))
+class RunResults extends Component {
+  render() {
+    const { selectedPrivateController, privateControllers } = this.props;
+    const selectedPrivateControllerData = privateControllers.filter(
+      score => score.result[0] === selectedPrivateController
+    );
+    const acceptanceStatus = selectedPrivateControllerData[0].seen;
```
gurbirkalsi: I'm receiving the following error here:

TypeError: Cannot read property 'seen' of undefined
RunResults.render
src/pages/RunResults/index.js:25
```
  22 | const selectedPrivateControllerData = privateControllers.filter(
  23 |   score => score.result[0] === selectedPrivateController
  24 | );
> 25 | const acceptanceStatus = selectedPrivateControllerData[0].seen;
     | ^  26 | 
  27 | const unAcceptedHint = (
  28 |   <Hint className={styles.customUnAccepteddHint}>
```

src/pages/RunResults/index.js
```diff
@@ -0,0 +1,71 @@
...
+  FlexItem,
+} from '@patternfly/react-core';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
+import styles from './index.less';
+
+@connect(({ privatedatastore }) => ({
+  selectedPrivateController: privatedatastore.selectedPrivateController,
+  privateControllers: privatedatastore.privateControllers,
+}))
+class RunResults extends Component {
```
gurbirkalsi: What if we recycled the `Summary` page component in order to show details regarding a result the user has selected? We can enhance the page component to allow users to take action on accepting/deleting the result. This would improve component reuse and keep "detail" views in one component.

mock/overview.js
```diff
@@ -0,0 +1,106 @@
+export const expirationLimit = 90;
+
+// P.S: the key needs to be unique for this object
+export const resultData = [
```
gurbirkalsi: What if we use a generator here to create an array of object literals? For example:
```
new Array(10).fill().map((value, index) => {
    {
      key: index,
      result: [`mock result - ${index}`, 'example.lab.eng.com'],
      end: '2021-02-10T 11:52:34.552478',
      deletion: '2021-04-20T11:52:34.552478',
      seen: false,
      saved: false,
      status: 'published',
      description: 'More content can be added here',
    },
})
```

mock/overview.js
```diff
@@ -0,0 +1,106 @@
...
+    end: '2021-01-20T11:52:34.552478',
+    deletion: '2021-04-20T11:52:34.552478',
+    seen: true,
+    saved: true,
+    status: 'unpublished',
+    description: 'More content can be added here',
+  },
+];
+
+export const expiringSoonResults = {
```
gurbirkalsi: See above comment on utilizing generators to create mock data. This will allow us to tune the amount of mock data we might need in the future by changing the number of items we specify for the array.

config/router.config.js
```diff
@@ -74,12 +74,27 @@ module.exports = [
...
+            component: './Overview',
+          },
           {
             path: '/private/profile',
             name: 'profile',
             exact: true,
             component: './Profile',
           },
+          {
+            path: '/private/runresults',
```
gurbirkalsi: We should consider simplifying the routing naming/schema to make the page's intention more clear to the user. What are your thoughts on the following schema?
```
[
... // other routes
{
  path: '/results',
  name: 'Results',
  component: './Results',
},
{
  path: '/results/new',
  name: 'new',
  component: './NewResults',
},
{
  path: '/results/expiring',
  name: 'expiring',
  component: './ExpiringResults',
}
... // other routes
]
```

This schema also suggests we would rename our other `/results` endpoint to `/result-selection` in order to differentiate the routes.

src/components/LoginForm/index.js
```diff
@@ -0,0 +1,145 @@
...
+  Title,
+} from '@patternfly/react-core';
+import { routerRedux } from 'dva/router';
+import { connect } from 'dva';
+import styles from './index.less';
+
+@connect(auth => ({
+  auth: auth.auth,
+}))
+class LoginForm extends React.Component {
```
gurbirkalsi: Why do we render a Login modal as opposed to redirecting the user to the Login page? If the Login modal is required, can we consider integrating the Login page component in order to improve component reuse?

src/components/LoginModal/index.js
```diff
@@ -0,0 +1,96 @@
...
+    const { dispatch } = this.props;
+    this.setState(({ isModalOpen }) => ({
+      isModalOpen: !isModalOpen,
+    }));
+    dispatch(routerRedux.push(`/`));
+  };
+
+  handleLoginModal = () => {
+    this.setState({
+      modalContent: <LoginForm />,
```
gurbirkalsi: We should consider an approach that doesn't require storing markdown in the component state in order to keep our component state clean & minimal in size. We can use conditional rendering here in order to render `<LoginForm />` depending on the user state. 

For more on conditional rendering: https://reactjs.org/docs/conditional-rendering.html

src/models/global.js
```diff
@@ -11,6 +11,8 @@ export default {
     selectedControllers: [],
     selectedFields: [],
     selectedIterations: [],
+    privateControllers: [],
```
gurbirkalsi: We have duplicate declarations for the `privateController` and `selectedPrivateController` fields in more than one Redux namespace.

src/models/privatedatastore.js
```diff
@@ -0,0 +1,38 @@
+export default {
```
gurbirkalsi: How is the `privatedatastore` namespace being used? What constitutes a private controller? We should be able to use the `global` namespace to track user state regardless of whether the user is authenticated.

src/models/user.js
```diff
@@ -4,6 +4,7 @@ export default {
   state: {
     favoriteControllers: [],
     favoriteResults: [],
+    seenControllers: [],
```
gurbirkalsi: Are we tracking controllers the user has seen or results? In the Overview page we specify "New and unmanaged Runs" and "Saved Runs".

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,204 @@
...
+        title: 'End Time',
+        dataIndex: 'end',
+        key: 'end',
+      },
+      {
+        title: 'Scheduled for deletion on',
+        dataIndex: 'deletion',
+        key: 'deletion',
+        render: text => {
+          const deleteDate = moment(new Date(Date.parse(text)));
```
gurbirkalsi: I believe you created a utility function to handle generating the date text that we can include a reference to here.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,204 @@
...
+          );
+        },
+      },
+      {
+        title: '',
+        dataIndex: 'action',
+        key: 'action',
+        render: (value, row) => {
+          return (
+            <div>
```
gurbirkalsi, line 252: Take a look at the `Menu` component from PatternFly for handling table actions.

src/pages/LoginHandler/index.js
```diff
@@ -1,146 +1,11 @@
...
-                Submit
-              </Title>
-            </Button>
-          </ActionGroup>
-        </Form>
-      </div>
-    );
-    return <AuthLayout toPreview={form} heading="Log into your Pbench Acount" backOpt="true" />;
-  }
+function LoginHandler() {
```
gurbirkalsi, line 5: This page component isn't being referenced in the codebase. Do we need this component definition?

gurbirkalsi (changes requested):

src/pages/Overview/index.js
```diff
@@ -1,16 +1,622 @@
...
+    const { newData, unlabledData, isOpen } = this.state;
+    const { favoriteControllers, seenControllers } = this.props;
+
+    const newDataColumns = [
+      {
+        title: 'Result',
+        dataIndex: 'result',
+        key: 'result',
+        render: (text, row) => {
+          let isSeen = false;
```
gurbirkalsi: I'm getting the following error here:

TypeError: Cannot read property 'forEach' of undefined
render
src/pages/Overview/index.js:212
```
  209 | render: (text, row) => {
  210 |   let isSeen = false;
  211 |   if (seenControllers !== []) {
> 212 |     seenControllers.forEach(item => {
      | ^  213 |       if (item.key === row.key) {
  214 |         isSeen = true;
  215 |       }
```

gurbirkalsi (changes requested):

src/pages/RunResults/index.js
anishaswain: @gurbirkalsi I am not being able to regenerate the above mentioned error. There might be issue with cache. Could you please check `privatedatastore` state in Redux DevTools? Or try opening accessing dashboard in incognito?

src/pages/RunResults/index.js
anishaswain: @gurbirkalsi that sounds great. I will mock the data for the recycled `Summary` page component for now and reuse the summary component here.

mock/overview.js
anishaswain: Fixed !

mock/overview.js
anishaswain: Fixed!

src/components/LoginModal/index.js
anishaswain: Fixed!

src/models/global.js
anishaswain: Removed.

src/models/privatedatastore.js
anishaswain: Moved selected private controllers by the user to `dashboard` namespace. I am using `privateControllers` for keeping track of the mocked controllers of user(will be replaced with `fetchControllers` in `dashboard` namespace later once we have data in server).

src/models/user.js
anishaswain: @gurbirkalsi we are tracking the results which the user has already seen/opened here. I changed the naming to `seenResults` mow.

src/pages/LoginHandler/index.js
anishaswain: @gurbirkalsi LoginHandler is the login page that we have in the dashboard now. It is referenced when the user hits `/login` URL.

src/pages/RunResults/index.js
anishaswain: Changed the namespace now. Please let me know if it works now.

src/pages/Overview/index.js
anishaswain: @gurbirkalsi could you please check the state in Redux DevTools? Or try opening accessing dashboard in incognito? I have changed the codebase now. Let me know if it works for you.

src/components/LoginForm/index.js
anishaswain: @gurbirkalsi I am not really sure why a login modal is required along with login page. It was discussed earlier in the design sync that it might enhance user experience. And the login modal itself reuses the `loginForm` but as discussed, we didn't want to include other login options and signup options in the modal so I added LoginForm component in the modal instead of Login Page component. Plus Login page component includes background image etc. which we do not want to be displayed in the modal.

src/pages/ExpiringResults/index.js
anishaswain: That's right. Used the same function now.

src/pages/Overview/index.js
anishaswain: waiting for #124 to get merged in order to avail PatternFly dropdown actions.

src/pages/ExpiringResults/index.js
anishaswain: waiting on #124 for incorporation PR table actions

config/router.config.js
anishaswain: @gurbirkalsi I think we should address the change of `/results` endpoint to `/result-selection` in another PR. To keep the routing naming changes seperate. What do you think?

config/router.config.js
```diff
@@ -74,12 +74,27 @@ module.exports = [
...
+            component: './Overview',
+          },
           {
             path: '/private/profile',
             name: 'profile',
             exact: true,
             component: './Profile',
           },
+          {
+            path: '/private/result',
```
gurbirkalsi: We may want to consider isolating the context of user authentication to the internals of the dashboard. For example, naming the route `/overview/result` in this case.

mock/api.js
```diff
@@ -29,11 +29,16 @@ export const mockResults = {
           controller_dir: casual.word,
         },
         run: {
+          key: index,
```
gurbirkalsi: What if we used the id field that contains a generated uuid as the key for rendering?

src/components/LoginModal/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+@connect(auth => ({
+  auth: auth.auth,
+}))
+class LoginModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isModalOpen: false,
+      modalView: 0,
```
gurbirkalsi: Can true and false be used for `modalView` state tracking?

src/components/LoginModal/index.js
```diff
@@ -0,0 +1,94 @@
...
+    const modalContent = modalView === 0 ? loginAction : <LoginForm />;
+    return (
+      <React.Fragment>
+        <Modal
+          variant={ModalVariant.small}
+          isOpen={isModalOpen}
+          onClose={this.handleModalCancel}
+          showClose="false"
+        >
+          {modalContent}
```
gurbirkalsi, line 87: Nice component reuse!

src/models/dashboard.js
```diff
@@ -35,29 +36,51 @@ export default {
...
     *fetchControllers({ payload }, { call, put }) {
       const controllers = yield call(queryControllers, payload);
-
+      console.log(controllers);
       yield put({
         type: 'getControllers',
         payload: controllers,
       });
     },
+    *fetchPrivateResults({ payload }, { put }) {
```
gurbirkalsi: What are your thoughts on removing the context of authentication and instead including ownership to naming of actions such as `fetchUserResults` in this case?

gurbirkalsi (dismissed):

mock/api.js
anishaswain: @gurbirkalsi we will need the `key` field for rendering. In the case of multiple selection, we are keeping track of `selectedRowKeys` which is an internal function requiring keys for each row. However I have fixed the same by assigning same value to `key` as `id`

src/components/LoginModal/index.js
anishaswain: fixed!

src/models/dashboard.js
anishaswain: The action was redundant so removed it. Thank you for noticing

config/router.config.js
anishaswain: @gurbirkalsi `/private` is added here because of the `private` wrapper inside which the routes are present. We won't be able to access the route without add `/private`  as a prefix to it. Either we need to delete the wrapper and handle the private vs public routes in some other way or rename the wrapper itself.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,726 @@
...
+      dropdownElem.style.display = 'block';
+    } else {
+      dropdownElem.style.display = 'none';
+    }
+  };
+
+  saveRuns = keys => {
+    const { totalResultData } = this.state;
+    const { dispatch } = this.props;
+    keys.forEach(function(key) {
```
aquibbaig: When there are no runs at first, clicking on save runs says "TypeError: Cannot read property 'forEach' of undefined"

aquibbaig (commented):
Can we break `src/pages/Overview/index.js` into smaller components? The whole file is about 700+ lines

src/pages/Overview/index.js
anishaswain: @aquibbaig please recheck if you are still facing the same issue

src/pages/Overview/index.js
```diff
@@ -1,16 +1,680 @@
-import React, { Component } from 'react';
-// import { routerRedux } from 'dva/router';
+import React from 'react';
```
gurbirkalsi, line 1: Seeing the following error on Overview page load: 

```
TypeError: Cannot read property 'forEach' of undefined
Cell
./src/pages/Overview/index.js:518
  515 | const row = cell.row.original;
  516 | let isSeen = false;
  517 | if (seenResults !== []) {
> 518 |   seenResults.forEach(item => {
      | ^  519 |     if (item.original.key === row.key) {
  520 |       isSeen = true;
  521 |     }
```

config/router.config.js
```diff
@@ -74,6 +74,16 @@ module.exports = [
         name: 'profile',
         component: './Profile',
       },
+      {
+        path: '/result',
+        name: 'result',
+        component: './Result',
+      },
+      {
+        path: '/expiringresults',
```
gurbirkalsi: Small nit: what if we named the route `expiring-results` to improve readability?

mock/api.js
```diff
@@ -297,7 +282,11 @@ export const mockSession = {
 export default {
   'GET /controllers/months': mockIndices,
   'POST /controllers/list': mockControllers,
-  'POST /datasets/list': mockResults,
+  'POST /datasets/list': (req, res) => {
```
gurbirkalsi: What if we moved this to a helper function to wrap the mocked results in an object? There may be cases in the future where this is required for other mocked responses.

src/components/LoginForm/index.js
```diff
@@ -0,0 +1,126 @@
...
+    dispatch({
+      type: 'auth/loginUser',
+      payload: {
+        username,
+        password,
+      },
+    });
+  };
+
+  /* eslint-disable no-restricted-syntax */
```
gurbirkalsi, line 45: We can remove this eslint rule comment.

src/components/RowSelection/index.js
```diff
@@ -1,22 +1,59 @@
...
+import CaretDownIcon from '@patternfly/react-icons/dist/js/icons/caret-down-icon';
 import Button from '../Button';
 
-export default class RowSelection extends PureComponent {
-  static propTypes = {
-    selectedItems: PropTypes.array.isRequired,
-    compareActionName: PropTypes.string.isRequired,
-    onCompare: PropTypes.func.isRequired,
-    style: PropTypes.object,
+export default class RowSelection extends React.Component {
```
gurbirkalsi, line 6: We should consider moving the "Manage runs" dropdown button to a separate component that is then reference in the `RowSelection` component. This will improve & help separate logic of both components here.

src/components/RowSelection/index.js
```diff
@@ -27,6 +64,22 @@ export default class RowSelection extends PureComponent {
...
           style={{ marginRight: 8 }}
         />
+        <Dropdown
+          onSelect={this.onSelectManageRunDropdown}
+          toggle={
+            <DropdownToggle
+              onToggle={this.onToggleManageRunDropdown}
+              toggleIndicator={CaretDownIcon}
+              isPrimary
+              id="toggle-id-4"
```
gurbirkalsi: We can remove the id prop for this component if not required by PatternFly.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,295 @@
...
+  ProgressVariant,
+  Tooltip,
+} from '@patternfly/react-core';
+import { Icon } from 'antd';
+import { EllipsisVIcon } from '@patternfly/react-icons';
+import { formatDate, getDiffDays, getDiffDate } from '../../utils/moment_constants';
+import Table from '@/components/Table';
+import styles from './index.less';
+
+const expirationLimit = { val: 30 };
```
gurbirkalsi: Does this need to be an object or can we simply assigned it to 30? Additionally, constants should be all uppercase (e.g. `EXPIRATION_LIMIT`) for improved readability.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,295 @@
...
+    const { dispatch, selectedControllers, selectedDateRange } = this.props;
+    dispatch({
+      type: 'dashboard/fetchResults',
+      payload: {
+        selectedDateRange,
+        controller: selectedControllers,
+      },
+    }).then(() => {
+      const { results } = this.props;
+      console.log(results[selectedControllers[0]]);
```
gurbirkalsi: `console.log` should not be included in production code.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,295 @@
...
+    console.log(controller);
+    const { dispatch } = this.props;
+    dispatch({
+      type: 'user/removeResultFromSeen',
+      payload: controller,
+    });
+  };
+
+  showDrowpdown = id => {
+    const dropdownElem = document.getElementById(id);
```
gurbirkalsi, line 116: We should avoid using the DOM API and use PatternFly's exposed API for displaying and hiding the dropdown menu.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,295 @@
...
+      },
+      {
+        Header: 'Scheduled for deletion on',
+        accessor: 'deletion',
+        Cell: cell => {
+          const remainingDays = getDiffDays(cell.value);
+          if (remainingDays > 15) {
+            return (
+              <div>
+                <Tooltip content={formatDate('without time', cell.value)}>
```
gurbirkalsi: We can move the conditional further into the markdown to avoid repeating code. For example: 

```
          return (
              <div>{remainingDays > 15 ? (
                <Tooltip content={formatDate('without time', cell.value)}>
<span>{getDiffDate(cell.value)}</span>
                </Tooltip>) : (<span>{cell.value}</span>)}
                <Progress
                  min={0}
                  max={expirationLimit.val}
                  value={expirationLimit.val - remainingDays}
                  size={ProgressSize.sm}
                  measureLocation={ProgressMeasureLocation.none}
                />
              </div>
            );
          }
```

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,295 @@
...
+          <TextContent className={styles.paddingSmall}>
+            <Text component={TextVariants.h1}> All runs</Text>
+          </TextContent>
+          <Grid hasGutter style={{ marginTop: '16px' }}>
+            <GridItem span={12}>
+              <Card>
+                <div>
+                  <Table
+                    columns={seenDataColumns}
+                    // expandedRowRender={record => <p style={{ margin: 0 }}>{record.description}</p>}
```
gurbirkalsi: We can remove this comment.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,680 @@
...
+        selectedDateRange,
+        controller: selectedControllers,
+      },
+    }).then(() => {
+      const { results } = this.props;
+      this.getSeperatedResults(results[selectedControllers[0]]);
+    });
+  }
+
+  getSeperatedResults(data) {
```
gurbirkalsi: Spelling: `getSeparatedResults`

src/pages/Overview/index.js
```diff
@@ -1,16 +1,680 @@
...
+    const unlabledData = data.filter(x => x.saved === true);
```
gurbirkalsi: Spelling: `unlabeledData`

src/pages/Overview/index.js
anishaswain: This is essentially caching issue. Can not re-generate the error.

config/router.config.js
anishaswain: fixed!

mock/api.js
anishaswain: fixed!

src/components/LoginForm/index.js
anishaswain: done

src/components/RowSelection/index.js
anishaswain: Yes, but then it would cause too much of props drilling.

src/components/RowSelection/index.js
anishaswain: removed!

src/pages/ExpiringResults/index.js
anishaswain: We need to discuss whether the value is constant or not and if not, where is the value stored. @portante @dbutenhof

src/pages/ExpiringResults/index.js
anishaswain: Since we aren't using PatternFly's table, we can't replicate the actions column used in PF tables. Since the dropdowns are created dynamically, it's difficult to use `refs` to correctly address them. @aquibbaig any suggestions on how to fix this issue without direct dom manipulation

src/pages/ExpiringResults/index.js
anishaswain: done !

src/pages/ExpiringResults/index.js
anishaswain: removed!

src/pages/Overview/index.js
anishaswain: fixed!

src/pages/Overview/index.js
anishaswain: fixed!

src/pages/ExpiringResults/index.js
dbutenhof: This is an attribute we're going to want to maintain on the server. In fact I suspect we'll have a tier of settings, with server defaults (and possibly maximums) for private and shared values, plus override values for each user. The latter is probably user metadata, which will be enabled by Nikhil's #2049; I'm not quite sure how we maintain the server defaults and limits. We probably need a new configuration API to get/set pre-defined values that aren't associated with any user or dataset, and we haven't talked about those. (Alternately, I suppose, these could be `pbench-server.cfg` values, although I like the idea of opening some class of configuration settings to more dynamic management through a hypothetical future dashboard "server management" page.)

For now, a hardcoded dashboard value is probably fine, but let's not treat it as a "constant" because in the long run, it won't be. Creating a redux store for "server configuration" is probably a good idea, and it can for now be hardcoded there until we have the server APIs built.

mock/augumented.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const EXPIRATION_LIMIT = { val: 30 };
+
+export const getUserResults = mockData => {
+  const augumentedData = mockData.map(item => ({
+    ...item,
+    deletion: moment.utc().add('days', 30),
+    seen: false,
+    saved: Math.random() < 0.5,
+    status: 'published',
```
dbutenhof: I'd prefer to use the server term, `access`, here. I like the consistency between "public" and "private", which are the current possible values; and we hope to expand this in the future to support generalized groups of users.

In any case, the `serverMetadata` block I'm going to return from `/datasets/list` and `datasets/detail` will use the term "access" ... although of course the dashboard is free to store that any way you see fit.

And why is `saved` randomly generated, but `seen` is always false? As a mock, it doesn't really make much difference, but logically  doesn't it have to be "seen" in order to be "saved"??

For reference on the server payload I'm going to be returning (from datasets/list as well as datasets/detail), see pbench issue [#2306](https://github.com/distributed-system-analysis/pbench/issues/2406).

You can control the metadata returned by requesting it with the new "metadata" request parameter: for example,

```
{
    "user": "brian@example.com",
    "controller": "brian",
    "start": "2021-06-01",
    "end": "2021-09-30",
    "metadata": ["deletion", "seen", "saved", "access", "owner"]
}
```

The idea is that the client may not need metadata (you can ask for it from the list API, the details API, both, or neither), and the set of allowed metadata can be expanded without affecting existing client responses by adding keywords later.

Technically, the value of each of these is JSON, although in practice right now they're all simple boolean or string values rather than JSON objects. I expect we'll want a "user" metadata object that can contain user-specified metadata tags; but we'll worry about the semantics (and UI) later.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,642 @@
...
+        totalResultData: data,
+      },
+      () => this.getSeparatedResults()
+    );
+  }
+
+  getSeparatedResults() {
+    const { totalResultData } = this.state;
+    const unlabeledData = totalResultData.filter(x => x.saved === true);
+    const newData = totalResultData.filter(x => x.saved !== true);
```
dbutenhof: So `unlabledData` is *saved* and `newData` is *unsaved*? Why `unlabled` instead of something like `savedData`?

It's also worth repeating here that I shifted the new metadata down a layer into "serverMetadata" to make sure we can't have any name collisions with run/metadata fields returned by the queries. So that will be `x.serverMetadata.saved`.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,642 @@
...
+      payload: totalResultData,
+    }).then(() => {
+      this.setState({
+        totalResultData,
+      });
+      this.getSeparatedResults();
+    });
+  };
+
+  deleteResult = (e, rows) => {
```
dbutenhof, line 180: Is this actually to *delete* a server dataset? That's not implemented yet in the server, but will be soon.

We'll also need an mechanism on the UI to publish a dataset. (And the API allows "unpublishing" a published dataset that you own, because you specify the desired "access" value -- "public" or "private".)

src/pages/Overview/index.js
```diff
@@ -1,16 +1,642 @@
...
+              ) : (
+                <span>{cell.value}</span>
+              )}
+              <Progress
+                min={0}
+                max={EXPIRATION_LIMIT.val}
+                value={EXPIRATION_LIMIT.val - remainingDays}
+                size={ProgressSize.sm}
+                measureLocation={ProgressMeasureLocation.none}
+              />
```
dbutenhof: A progress bar? This seems a bit odd, since you seem to be showing the "progress" of the dataset's expiration window relative to the arbitrary maximum allowed, which doesn't really seem all that useful to me.

And using the `EXPIRATION_LIMIT` literal from your mock spreads knowledge of the mock further than necessary through the code. Can't we isolate this in Redux where we can seamlessly transition from built-in mocks to real data when it's available, without having to find all the places that were faked?

I anticipate two different "expiration limits", at the server level (controlled by the server administrator) and at the user level, which would appear as user metadata. (We don't have either yet, unfortunately, although Nikhil's user metadata PR provides the mechanism for the latter just as my dataset metadata provides the basis for your `SEEN` and `SAVED` metadata.)

We should expand the user context in Redux to include a per-user maximum expiration, which can be displayed and (eventually) edited as part of the user profile. This would be limited (in the server) against a server maximum value controlled by the server administrator. (I'd also eventually like to have a server administration console in the dashboard that allows an ADMIN user to interact directly with those, but we're not nearly there yet.)

Anyway, the point is that if you mocked things like this cleanly within a Redux context we won't have to touch so much code to roll out the server feature; and the dashboard code will look more consistent.

dbutenhof (commented):
I haven't completely absorbed all this yet, but I realized that I wrote up a comment last week that I forgot to submit; so I added a few more and I'm going to commit the set of comments so you can see them. I'll make another pass later.

src/models/dashboard.js
```diff
@@ -44,6 +45,11 @@ export default {
     *fetchResults({ payload }, { call, put }) {
       const results = yield call(queryResults, payload);
 
+      // post processing to add additional data to results
+      const controller = payload.controller[0] || 'mock-controller';
+      const augumentedResults = getUserResults(results[controller]);
```
dbutenhof: I don't think the main dashboard logic should be referencing anything under the umi `mock` directory; those should be transparent testing infrastructure. Are these even included in the build? Also, with the way you're using it, there's no point in putting this under the `mock` directory at all, since it's not being activated "automatically" by the test framework. The code could be right here at the model layer.

The advantage in that would be that your code can run against the real server with faked metadata until we implement the data in the server; however at that point we'll have to revert the dashboard "augmentation" to work with the real server data.

The alternative would be to just change the result mock in `mock/api.js` so in your development server testing, you're using the "future server" data; although this means it won't work against the current real server.

portante (changes requested):
I personally need a walk through on the changes to help me understand everything.  In the mean time, it looks like the PR needs a rebase.

src/components/LoginForm/index.js
```diff
@@ -0,0 +1,126 @@
...
+  const [username, setUsername] = useState('');
+  const [password, setPassword] = useState('');
+  const [errors, setErrors] = useState({
+    email: '',
+  });
+  const [btnDisabled, setBtnDisabled] = useState(true);
+
+  const handleUsernameChange = val => {
+    setUsername(val);
+    const validEmail = validateEmail(val);
```
dbutenhof: I'd like to drop this validation. Right now the dashboard is validating "username as email" on the registration form; here, we shouldn't really need to care as long as the given username is in the server's IAM. We want to reconcile the current server and dashboard user model by adding a real "username" in addition to the email; but right now you can use the API (e.g., through curl) or a server command to create a user with a non-email username ... and this "validation" prevents such a user from being logged in.

This isn't the topic of this PR, so I won't object on this basis, but since you're moving the code around this may be as good a time as any for that one small change. (In any case, we don't need to address the dashboard user model change here; I put in a separate issue for that.)

src/components/LoginForm/index.js
```diff
@@ -0,0 +1,126 @@
...
+        setBtnDisabled(false);
+      } else setBtnDisabled(true);
+    },
+    [username, password]
+  );
+
+  const form = (
+    <div className={styles.section}>
+      <Form className={styles.section}>
+        <FormGroup label="Email address" isRequired fieldId="horizontal-form-name">
```
dbutenhof: Changing "Email address" to "Username" would be good, too; but again, not essential for this PR.

src/components/RowSelection/index.js
```diff
@@ -1,22 +1,59 @@
...
+    const manageRunDropdown = [
+      <DropdownItem key="save" onClick={saveRuns}>
+        {' '}
+        Save Runs
+      </DropdownItem>,
+      <DropdownItem key="unread" onClick={removeResultFromSeen}>
+        {' '}
+        Mark as unread
+      </DropdownItem>,
+      <DropdownItem key="favourite" onClick={favoriteResult}>
```
dbutenhof: It's interesting that the "key" uses the British English spelling while the onClick function uses the American English spelling. Consistency would be better here, one way or the other.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,343 @@
...
+    dispatch({
+      type: 'dashboard/fetchResults',
+      payload: {
+        selectedDateRange,
+        controller: selectedControllers,
+      },
+    }).then(() => {
+      const { results } = this.props;
+      this.setState({
+        totalResultData: results[selectedControllers[0] || 'mock-controller'],
```
dbutenhof, line 58: Is `'mock-controller'` something we want in "production" dashboard code? I'm not sure from the logic here exactly when or how we might have `selectedControllers[0]` as falsy, but it doesn't seem we ever want to go to `mock-controller` (whatever that'll actually do) except in testing.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,343 @@
...
+
+  deleteResult = (e, rows) => {
+    // Stop propagation from going to the next page
+    if (e !== null) {
+      e.stopPropagation();
+    }
+    const { totalResultData } = this.state;
+    const { dispatch } = this.props;
+    const updatedResult = totalResultData.filter(item => !rows.includes(item.result));
+    dispatch({ type: 'dashboard/updateResults', payload: updatedResult })
```
dbutenhof, line 89: Interesting. You have `user/removeResultFromFavorites` which, eventually, will be able to make an API call to modify the per-user metadata to manage server-based favorites; and `user/removeResultFromSeen` which can use the dataset metadata to manage the `dashboard.seen` property; but here we only `dashboard/updateResults` which I don't think can practically make an API call to actually delete a dataset. I assume it's only updating the Redux state, which means the "deleted" dataset will appear again on a refresh?

But this also raises an interesting meta-question: I planned on managing seen/saved as dataset metadata, because it's only really relevant (I assumed) for the *owner* of the data, and it makes sense to track this as a property of the dataset. Doing this through the `user` Redux state implies it's a per-user property as in, anyone viewing a public dataset owned by someone else might mark it "seen" or "unseen". That's not the way the overview page is working, and I'm not sure it makes much sense... which suggests that "user" probably isn't the right place for managing seen/saved.

But I suppose we can work all this out later.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,343 @@
...
+
+  removeResultFromSeen = row => {
+    const { dispatch } = this.props;
+    dispatch({
+      type: 'user/removeResultFromSeen',
+      payload: row,
+    });
+  };
+
+  showDrowpdown = (e, id) => {
```
dbutenhof: Typo: "drowpdown" 😀

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,343 @@
...
+          pathname: '/result',
+        })
+      );
+    });
+  };
+
+  markResultSeen = row => {
+    const { dispatch } = this.props;
+    dispatch({
+      type: 'user/markResultSeen',
```
dbutenhof, line 142: Again, it doesn't seem that "seen" should be considered "user" state, but the packaging is of secondary importance here and we can work that out later when we integrate with the server APIs.

dbutenhof (dismissed):
I have a bunch of comments with small changes and architectural issues that don't necessarily need to be resolved in this PR; we'll need another follow-on PR to integrate this work with the server APIs in any case.

src/models/dashboard.js
dbutenhof: This is funny, because I don't remember writing it. I think I had this already "pending" from a previous review long, long, ago, and I forgot to post it ... and then it came along with today's review comments. Although I suspect it's still mostly true, and I'm not really sure what to think about the logical-or with the `mock-controller` name, which definitely seems out of place here.

webbnh: Just so I'm understanding...it's not the "logical-or" itself which is the issue, it's the application of the `mock-controller` name as a _default_ in this code...right?

dbutenhof: I'm going to just resolve this old comment; I made a comment yesterday on the `|| 'mock-controller'` phrase and the rest of this comment is a bit unfocused.

src/pages/Overview/index.js
```diff
@@ -1,16 +1,850 @@
...
+    if (expiringData.length > 0) {
+      this.setState({
+        expiringData: [],
+      });
+    } else {
+      this.setState({
+        expiringData: newData.filter(x => getDiffDays(x.serverMetadata['server.deletion']) < 15),
+      });
+    }
+  }
```
webbnh: First, LGTM is flagging line 249 with the warning, "Component state update uses potentially inconsistent value."  It would be good to understand (and hopefully fix!) that before merging.

That noted, I'm having trouble with this code:  if I understand what I'm reading, then if there is expiring data then we ignore the `newData` and clear the existing expiring data; however, if there is no expiring data, then we set the expiring data based on the new data.  Is that right?  I would naively have thought that we would always set the expiring data based on the new data, and, if there is no new data, then we would clear the expiring data.

Finally, the `15` is a magic number which really should be defined as a symbolic constant/variable.  That will make it easy to change or to make it dynamic in the future.

src/pages/ExpiringResults/index.js
portante: I am not sure it makes sense to have "seen"/"unseen" be for public data sets.  That is intended as part of a user's data curation work flow.

If we want to allow another user to "favorite", or mark as "seen"/"unseen" a public data set, we should implement that separate from data curation.

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,343 @@
...
+      },
+      {
+        Header: '',
+        accessor: 'action',
+        Cell: cell => {
+          const row = cell.row.original;
+          return (
+            <div>
+              <EllipsisVIcon
+                onClick={e => this.showDrowpdown(e, `newrun${row.result}`)}
```
portante: At least it was used consistently... ;)

src/pages/Overview/index.js
```diff
@@ -1,16 +1,850 @@
...
+    if (expanded) {
+      this.setState({ expanded: false });
+    } else {
+      this.setState({ expanded: true });
+    }
+  }
+
+  getRunResult() {
+    const { results, selectedControllers } = this.props;
+    const data = results[selectedControllers[0] || 'mock-controller'];
```
portante, line 74: Another case of the `mock-controller` showing up in non-test code.

portante (dismissed):
I don't have much to add beyond what Dave pointed out already.  Aside from the spelling of a function, and the references to `mock-controller` in the non-test code, it looks good to go.

src/pages/ExpiringResults/index.js
webbnh: Otherwise there would have been a reference error!  😀

src/components/LoginForm/index.js
anishaswain: removed for now!

src/components/LoginForm/index.js
anishaswain: changed!

src/components/RowSelection/index.js
anishaswain: Fixed! This was particularly interesting!

src/pages/ExpiringResults/index.js
anishaswain: In overview page, we don't have a concept of `controller` like in result page. And in-order to use result query we need a `selectedController` name which is `undefined` in case of overview page. So if we don't provide `mock-controller` , UI will fail.

src/pages/ExpiringResults/index.js
anishaswain: fixed!

src/pages/Overview/index.js
anishaswain: @webbnh I have taken this approach to show demo of having and not having expiring data. You are right. It would depend only on new data. I have removed the demo code now. 

Also, yeah added `15` as a constant value.

src/pages/ExpiringResults/index.js
anishaswain: sorry! :D

fixed it now.

src/pages/Overview/index.js
anishaswain: In overview page, we don't have a concept of `controller` like in result page. And in-order to use result query we need a `selectedController` name which is `undefined` in case of overview page. So if we don't provide `mock-controller` , UI will fail.

src/pages/Overview/index.js
```diff
@@ -31,6 +30,8 @@ import { formatDate, getDiffDays, getDiffDate } from '../../utils/moment_constan
 import Table from '@/components/Table';
 import styles from './index.less';
 
+const expiringLimit = 15;
```
dbutenhof, line 33: This is better than an inline literal. I'm inclined to let this go for now in favor of getting the basic Overview structure into `main`; we will however want to move this out into Redux, because I think we're going to want this to be a user preference setting at some point.

src/pages/ExpiringResults/index.js
dbutenhof: This is really interesting. We've talked about making the concept of "controller" less important, for a number of reasons; including the idea that it was initially a tool to separate datasets for users who each generally had their own test hostname... whereas now we have real users *and* many of our common test environments have transient hostnames that aren't really useful.

Right now the server queries are organized, like the original dashboard services, to search on controller and name, but maybe we should consider removing controller or at least making that parameter optional. As long as our dataset names are unique -- and some legacy server code seems to already assume and depend on that -- removing the controller term shouldn't have a big impact.

Then you'd feed this search simply with the `/controllers/months` date range (which is "everything", since there's no date range selector on the Overview page), and the logged-in user.

For right now, though, what does this artifact actually mean? Under what circumstances will we have a `selectedController` when we run this against a real server? Only if they've previously gone through the selection process on the main dashboard page; which I'm not sure makes much sense if the Overview page is the new main landing page.

Does that mean this PR is incomplete?

src/pages/ExpiringResults/index.js
dbutenhof: We don't currently have any workflow for a user other than the owner using saved/seen. Obviously we could invent one. I'd imagined that "favorite" would be maintained in per-user metadata, but as saved/seen are part of the *owner's* dataset curation, they belonged with the dataset metadata model.

While we have the initial model for user metadata in Nikhil's [#2049](https://github.com/distributed-system-analysis/pbench/pull/2049), that was dropped since, at the time, we didn't have a strong commitment to any user story requiring per-user metadata. I've remained convinced we'll need this for things like persisted favorites, per-user default expiration time, probably dashboard visibility preferences, but completing the PR was put on the back burner in favor of more critical needs like authenticated queries.

Note that right now the dataset metadata reserves a "namespace" for the dashboard but doesn't define any specific data within that namespace. I envisioned `dashboard.saved`, for example, but we can work out an equivalent namespace model for per-user metadata if we want to put it there, along with favorites, etc. (Even something like `dashboard.saved.c301fa19de05031c1fab33f401acb0ae = True`)

dbutenhof (dismissed):

src/pages/Overview/index.js
webbnh: There are also three places where we are still using the literal instead of referencing a common symbol -- those should be symbolized as well, hopefully referring to the _same_ instance....

src/components/LoginForm/index.js
```diff
@@ -0,0 +1,123 @@
...
+            </Title>
+          </Button>
+        </ActionGroup>
+      </Form>
+    </div>
+  );
+  return <React.Fragment>{form}</React.Fragment>;
+};
+
+export default connect(mapStateToProps)(LoginForm);
```
aquibbaig, line 123: did we rename this component?

src/pages/ExpiringResults/index.js
aquibbaig: I think a valid question to ask here is "What happens when the user has no selected controllers in the redux store" or "totalResults" in the state is undefined

src/pages/Overview/index.js
```diff
@@ -1,16 +1,827 @@
...
+
+  saveRuns = (e, rows) => {
+    // Stop propagation from going to the next page
+    if (e !== null) {
+      e.stopPropagation();
+    }
+    const { totalResultData } = this.state;
+    const { dispatch } = this.props;
+    const keys = rows.map(({ original }) => original.result);
+    keys.forEach(function(key) {
```
aquibbaig: Minor fix recommended by eslint

keys.forEach((key) => {
    totalResultData.filter(item => item.result === key)[0].serverMetadata.dashboard.saved = true;
});

src/pages/Overview/index.js
```diff
@@ -1,16 +1,827 @@
...
-      <div>
-        <h1>Overview Page</h1>
+      <div className={styles.paddingBig}>
+        {actionMessage !== '' ? (
+          <AlertGroup isToast style={{ display: 'block' }}>
+            <Alert title={actionMessage} variant="success" />
+          </AlertGroup>
+        ) : (
+          <></>
+        )}
```
aquibbaig: we can go w `&&` for readability

src/pages/ExpiringResults/index.js
```diff
@@ -0,0 +1,343 @@
...
+          </AlertGroup>
+        ) : (
+          <></>
+        )}
+
+        <div className={styles.paddingBig}>
+          <TextContent className={styles.paddingSmall}>
+            <Text component={TextVariants.h1}> All runs</Text>
+          </TextContent>
+          <Grid hasGutter style={{ marginTop: '16px' }}>
```
aquibbaig, line 292: I think we can do some padding here, the search bar looks  a bit weird
demo

aquibbaig (dismissed):
Great work Anisha! 🎊

src/pages/Overview/index.js
portante: Tracked in issue #205.

src/pages/ExpiringResults/index.js
portante: We can track the need for a controller in issue #205.

src/components/LoginForm/index.js
anishaswain: No

src/pages/ExpiringResults/index.js
anishaswain: To be discussed further while doing API integration.

src/pages/Overview/index.js
anishaswain: fixed!

src/pages/Overview/index.js
anishaswain: changed!

src/pages/ExpiringResults/index.js
anishaswain: fixed with some margin

dbutenhof (approved):

aquibbaig (approved):

PR #106, opened 2020-12-01
Reference API endpoints via environment variables

gurbirkalsi:
This PR aims to relocate the `datastoreConfig.js` config file to `config/endpoints.js` and dynamically load with the umi config to expose environment variables globally once the Node.js process boots up.

This migration removes the need to query the endpoint config file within the page component mounts for each component referenced as a "base" route in the dashboard. This removes a layer of abstraction as API endpoints are now only referenced in the `services` directory for queries originating from the respective component. 

In order to support initialization of a Sentry instance at the earliest entrypoint of the dashboard codebase, a reference to the Sentry endpoint url needs to be made available directly after the Node.js process boots in order to optimize the scope of information available to collect at dashboard initialization. 

In addition, this change makes the transition to the singular pbench server API as simple as modifying `endpoint.js` and removing post processing where necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ _roadhog-api-doc
 .DS_Store
 npm-debug.log*
 yarn-error.log
-/mock/datastoreConfig.js
 
 /coverage
 .idea
@@ -30,3 +29,6 @@ package-lock.json
 
 # jest tests
 __snapshots__
+
+# config
+/config/endpoints.js
--- a/README.md
+++ b/README.md
@@ -8,10 +8,12 @@ Pbench Dashboard is a web-based platform for consuming indexed performance bench
 ├── public
 │   └── favicon.ico                 # favicon
 ├── mock
-│   └── datastoreConfig.js.example  # datastore configuration
+│   ├── api.js                      # mocked api
+│   └── user.js                     # mocked user api
 ├── config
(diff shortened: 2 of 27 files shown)


dbutenhof (approved):
> In addition, this change makes the transition to the singular pbench server API as simple as modifying endpoint.js and removing post processing where necessary.

More or less! ;-)

anishaswain (approved):
LGTM :+1:

portante (commented):
Not sure I quite understand all the functional changes of this PR.

So in the past, when I would go to `http://host.example.com/dashboard/`, looking at the Network console of my browser, I would see a load of `config.json` from the server, right?

Now UMI will build the contents of `endpoints.js` into the loaded java script, so at `yarn build` time we need to use the correct `endpoints.js` for the target server, right?

gurbirkalsi:
Correct, that network request for the config no longer exists and instead is now included as part of the webpack bundle initiated by Umi. I have a follow up PR in the works to update our Ansible scripts to generate the `endpoints.js` file that is included and served as part of the bundle deployed to the server.

portante:
This is a good interim step.

Let's consider a future step where the server would offer an end-point which would return the configuration relative to that server.  That way we can build one dashboard bundle and test it on staging and move it to production without having to change or rebuild the bundle.

PR #108, opened 2020-12-03
favorite records for registered users

### LINKED ISSUE #94: Unable to remove favorite records in the table

aquibbaig:
This PR introduces the following changes
- Hide the "actions" column for non-registered users, preventing them to bookmark records.
- Only registered users should be able to bookmark records and remove them.
- Fixes #94

### DIFF
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -25,9 +25,21 @@ export default {
         payload,
       });
     },
+    *removeControllerFromFavorites({ payload }, { put }) {
+      yield put({
+        type: 'removeFavoriteController',
+        payload,
+      });
+    },
     *favoriteResult({ payload }, { put }) {
       yield put({
         type: 'modifyFavoritedResults',
+        payload,
+      });
+    },
+    *removeResultFromFavorites({ payload }, { put }) {
+      yield put({
+        type: 'removeFavoriteResult',
         payload,
       });
     },
@@ -56,6 +68,18 @@ export default {
       return {
         ...state,
         favoriteResults: [...state.favoriteResults, payload],
+      };
+    },
+    removeFavoriteController(state, { payload }) {
+      return {
+        ...state,
+        favoriteControllers: state.favoriteControllers.filter(item => item.key !== payload.key),
+      };
+    },
+    removeFavoriteResult(state, { payload }) {
+      return {
+        ...state,
+        favoriteResults: state.favoriteResults.filter(item => item.key !== payload.key),
       };
     },
   },
--- a/src/pages/Controllers/index.js
+++ b/src/pages/Controllers/index.js
@@ -142,6 +142,17 @@ class Controllers extends Component {
     });
   };
 
+  removeControllerFromFavorites = (event, value, controller) => {
+    // Stop propagation from going to the next page
+    event.stopPropagation();
+    const { dispatch } = this.props;
+    // dispatch an action to favorite controller
+    dispatch({
+      type: 'user/removeControllerFromFavorites',
+      payload: controller,
+    });
+  };
+
   render() {
     const { controllers } = this.state;
     const { loadingControllers, favoriteControllers, auth } = this.props;
@@ -184,7 +195,11 @@ class Controllers extends Component {
             }
           });
           if (isFavorite) {
-            return <Icon type="star" theme="filled" />;
+            return (
+              <a onClick={e => this.removeControllerFromFavorites(e, null, row)}>
+                <Icon type="star" theme="filled" />
+              </a>
+            );
           }
           return (
             <a onClick={e => this.favoriteRecord(e, null, row)}>
--- a/src/pages/Controllers/index.js
+++ b/src/pages/Controllers/index.js
@@ -18,14 +18,17 @@ import SearchBar from '@/components/SearchBar';
 import AntdDatePicker from '@/components/DatePicker';
 import Table from '@/components/Table';
 import { getDiffDate } from '@/utils/moment_constants';
+import { isLoggedInUser } from '@/utils/utils';
+import styles from './index.less';
 
 const { TabPane } = Tabs;
 
-@connect(({ datastore, global, dashboard, loading, user }) => ({
+@connect(({ datastore, global, dashboard, loading, user, auth }) => ({
   controllers: dashboard.controllers,
   indices: datastore.indices,
   selectedDateRange: global.selectedDateRange,
   favoriteControllers: user.favoriteControllers,
+  auth: auth.auth,
   loadingControllers:
     loading.effects['dashboard/fetchControllers'] || loading.effects['datastore/fetchMonthIndices'],
 }))
@@ -141,7 +144,7 @@ class Controllers extends Component {
 
   render() {
     const { controllers } = this.state;
-    const { loadingControllers, favoriteControllers } = this.props;
+    const { loadingControllers, favoriteControllers, auth } = this.props;
     const columns = [
       {
         title: 'Controller',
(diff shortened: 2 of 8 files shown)


src/models/user.js
```diff
@@ -58,5 +70,17 @@ export default {
...
+        ...state,
+        favoriteControllers: state.favoriteControllers.filter(item => item.key !== payload.key),
+      };
+    },
+    removeFavoriteResult(state, { payload }) {
+      return {
+        ...state,
+        favoriteResults: state.favoriteResults.filter(item => item.key !== payload.key),
+      };
+    },
```
dbutenhof, line 84: This isn't directly related to your changes, but I hadn't really thought about this before. It's interesting that "favorites" is just a transient Redux property, and not persisted across browser sessions. Should this be part of user metadata at some point?

dbutenhof (approved):

src/pages/Controllers/index.js
```diff
@@ -139,9 +140,23 @@ class Controllers extends Component {
...
+      type: 'user/removeControllerFromFavorites',
+      payload: controller,
+    });
+  };
+
   render() {
     const { controllers } = this.state;
     const { loadingControllers, favoriteControllers } = this.props;
+    const {
+      auth: { username },
```
portante: So we are using the presence of a "username" field in the auth part of the Redux store?  Should this be a method so that we don't to change so many other modules the implementation changes over time?

riya-17 (approved):

src/models/user.js
aquibbaig: @dbutenhof Exactly, this is stored as in a "browser" storage, hence persisted across browser sessions. We also would like to have a functionality which lets users clean their local storage, which would delete all these records. To make it persist across "user" sessions, we would ideally save it in the backend smh. @portante Is there a need to do that in the near future?

src/pages/Controllers/index.js
aquibbaig: Are you suggesting this to be a method like `getUsername()` which returns the username once it exists, null otherwise? Either way, "auth" is returned to this component as a prop, so I believe there should not be any harm in directly accessing it.

Ideally, I would like the "user" to be in the global state, passed down to lower components in the hierarchy as a context object, this reduces the redundancy of accessing "username" in every page component. We would just get the user field directly after that.

portante: > Are you suggesting this to be a method like `getUsername()` which returns the username once it exists, null otherwise? Either way, "auth" is returned to this component as a prop, so I believe there should not be any harm in directly accessing it.

True, "no harm", but in the abstract the code is really trying to say: "A user is requesting that the Dashboard marks a controller as favorite.  Is this user logged in or not?"

Instead of depending on the property existing to infer the user is logged in or not, why not make a method that returns the answer to the question explicitly and directly?  Perhaps `isLoggedInUser()`?

Then we can define that method to always answer that question `true`/`false`, and over time if we have to change the implementation details, the method does not have to change.

> Ideally, I would like the "user" to be in the global state, passed down to lower components in the hierarchy as a context object, this reduces the redundancy of accessing "username" in every page component. We would just get the user field directly after that.

Sure, that is an implementation detail, but the method should hide the knowledge of how to use the global state to answer a question.  We provide a method to answer the question directly.

aquibbaig: Right! Got the point!

src/pages/Controllers/index.js
```diff
@@ -139,6 +140,28 @@ class Controllers extends Component {
     });
   };
 
+  isLoggedInUser = () => {
```
portante: Is there way to avoid repeating this code (DRY principle)?

portante (changes requested):

src/pages/Controllers/index.js
aquibbaig: Yes, there is! The way I can think of is using a global object/method that keeps record of whether a user is logged in and then accessing it directly in the child components? In our design, the component that can serve this is `BasicLayout`

src/global.less
```diff
@@ -29,3 +29,7 @@ body {
     margin: 40px 0 !important;
   }
 }
+
+.hide {
```
gurbirkalsi: Is this CSS being referenced somewhere? I didn't find any reference to it in markup.

src/pages/Controllers/index.js
gurbirkalsi: +1, have you also considered moving this to the `utils` directory as a simple helper function?

npalaska (approved):

src/global.less
portante: @aquibbaig ?

src/pages/Controllers/index.js
portante: @aquibbaig ?

src/models/user.js
portante: > We also would like to have a functionality which lets users clean their local storage, which would delete all these records.

Can you file an issue to track this?

> To make it persist across "user" sessions, we would ideally save it in the backend smh. @portante Is there a need to do that in the near future?

This will become a requirement once the notion of a user is implemented in both the Dashboard and in the backend server.

src/global.less
aquibbaig: Hey @gurbirkalsi. I use this css as a className in our table component definition to hide the favorites column for non-registered users

src/pages/Controllers/index.js
aquibbaig: Hey @gurbirkalsi @portante, yes! A simple helper function would help. Also, I probably need to convert BasicLayout to a functional component, if we do that we won't be able to use @connect decorator like we usually do. So, we need to move that to the end of the file where we export the component, something like a 'MapStateToProps' function would do.

src/global.less
gurbirkalsi: We might want to consider conditionally appending the `actions` column if the user is logged in during the render process. In general, CSS defined in `global.less` should be reserved for defining base layout/fonts.

src/pages/Controllers/index.js
gurbirkalsi: Why would BasicLayout need to be converted into a functional component to use the helper function? My impression is we would import the helper function into the relevant page component (in this case `Controllers/index.js`)

aquibbaig: Hey @gurbirkalsi, I actually tried using the `useContext` hook, which requires a functional component. The other way around is to use legacy context.

gurbirkalsi (approved):

gurbirkalsi:
@aquibbaig, seems we need a rebase in order to resolve the Travis CI errors.

src/global.less
gurbirkalsi: We can remove this CSS now that separate .less files have been created for each page component.

src/pages/Results/index.js
```diff
@@ -143,14 +146,24 @@ class Results extends Component {
     });
   };
 
+  removeControllerFromFavorites = (event, value, controller) => {
```
portante: This looks like almost identical code as found in `Controllers/index.js`, is it?

portante (changes requested):

src/pages/Results/index.js
aquibbaig: Nice catch!

portante (approved):

anishaswain (approved):

PR #110, opened 2020-12-07
Update Ansible configuration for deploying dashboard using runtime environment variables

gurbirkalsi:
The initialization of a `config.json.j2` file is no longer required as the dashboard bundles the endpoint configuration file (`config/endpoints.js`) into the final binary deployed to a remote host. In order to reflect this change, the `inventory` file has been simplified to only reference hosts being deployed to under the `[servers]` key.

The process for deploying the dashboard to a remote host using Ansible remains the same by running the following:
```
$ yarn build
$ ansible-playbook -i inventory dashboard-deploy.yml
```

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -3,20 +3,32 @@ This will ease installation, and deployment of the pbench dashboard.
 
 ## Required
 - Ansible needs to be installed on the host where you want to run this playbook
-- An inventory file containing the following key values defined:
-  - "`elasticsearch_url`", "`results_url`", "`graphql_url`", "`run_index`", "`prefix`"
-    See the `/README.md` for more details. 
+- An inventory file containing the server values defined
+
+## Endpoint Configuration
+API endpoints are defined in `public/endpoints.js` as runtime configuration variables for reference in the dashboard binary. Before deployment of the binary to a remote host, consider the configuration at `public/endpoints.js` as the file is copied to the target server during deployment. Please reference the following example for required configuration fields:
+
+```JavaScript
+window.endpoints = {
+  elasticsearch: 'http://test_domain.com',
+  results: 'http://test_domain.com',
+  graphql: 'http://test_domain.com',
+  prefix: 'test_prefix.',
+  result_index: 'test_index.',
+  run_index: 'test_index.'
+};
+```
 
 ## Run
 Running the below commands from this checked-out directory to install the
 pbench dashboard components locally, and then deploy hosts mentioned under
-the "`[servers:children]`" section of the given `inventory` file.
- 
-There's also an option to define the dashboard configuration in the provided
-inventory file.
+the "`[servers]`" section of the given `inventory` file.
 
 See the `inventory` file in this directory for an example.
 ```
+$ # First bundle the dashboard for production
+$ yarn build
+$
 $ # First add a link to the "dist" folder where the dashboard will be built.
 $ ln -sf ../dist dist
 $ ansible-playbook -i inventory dashboard-install.yml
--- a/package.json
+++ b/package.json
@@ -1,12 +1,11 @@
 {
   "name": "pbench_dashboard",
-  "version": "2.2.0",
+  "version": "2.2.1-0",
   "description": "UI solution for scalable visualization of benchmark data.",
   "private": true,
   "scripts": {
     "precommit": "npm run lint-staged",
-    "start": "cross-env UMI_UI=none umi dev",
-    "start:no-mock": "cross-env MOCK=none umi dev",
+    "start": "cross-env UMI_UI=none MOCK=none umi dev",
     "build": "npm --no-git-tag-version version prerelease && umi build",
     "site": "umi-api-doc static && gh-pages -d dist",
     "analyze": "cross-env ANALYZE=true umi build",
--- a/src/pages/document.ejs
+++ b/src/pages/document.ejs
@@ -6,7 +6,10 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Pbench Dashboard</title>
-  <link rel="icon" href="/favicon.ico" type="image/x-icon">
+  <link rel="icon" href="<%= context.config.publicPath +'favicon.ico'%>" type="image/x-icon">
+  
+  <!-- runtime endpoints config -->
+  <script src="<%= context.config.publicPath +'endpoints.js'%>"></script>
 </head>
 
 <body>
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,4 @@ package-lock.json
 __snapshots__
 
 # config
-/config/endpoints.js
+public/endpoints.js
--- a/README.md
+++ b/README.md
@@ -70,17 +70,17 @@ This will automatically open the application on [http://localhost:8000](http://l
 
 Both the production and development builds of the dashboard require API endpoint configurations in order to query data from specific datastores.
 
-`endpoints.js` in the `config/` directory contains references to datastores required to visualize data in the dashboard. Please reference the following example for required configuration fields.
+`public/endpoints.js` contains references to datastores required to visualize data in the dashboard. Please reference the following example for required configuration fields:
 
 ```JavaScript
-export default {
-  "elasticsearch": "http://elasticsearch.example.com",
-  "results": "http://results.example.com",
-  "graphql": "http://graphql.example.com",
-  "prefix": "example.prefix",
-  "run_index": "example.index",
-  "result_index": "example.index"
-}
+window.endpoints = {
+  elasticsearch: 'http://test_domain.com',
+  results: 'http://test_domain.com',
+  graphql: 'http://test_domain.com',
+  prefix: 'test_prefix.',
+  result_index: 'test_index.',
+  run_index: 'test_index.'
+};
 ```
 
 ## Storage Config
--- a/ansible/config.json.j2
+++ b/ansible/config.json.j2
@@ -1,1 +1,0 @@
-../config.json.j2
--- a/ansible/inventory
+++ b/ansible/inventory
@@ -1,22 +1,3 @@
-[servers:children]
-ui-server
-user-server
-
-[servers:vars]
-results_url="http://{{ inventory_hostname }}"
-run_index="run.example"
-graphql_url="http://{{ inventory_hostname }}"
-
-[ui-server]
-ui-server.example.com
-
-[ui-server:vars]
-elasticsearch_url="http://elasticsearch.example.com"
-prefix="ui-server.example."
-
-[user-server]
-user-server.example.com
-
-[user-server:vars]
-elasticsearch_url="http://elasticsearch.example.com"
-prefix="user-server.example."
+[servers]
+staging-server.example.com
+production-server.example.com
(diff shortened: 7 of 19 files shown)


portante (changes requested):
But how does the user verify that the `yarn build` will work with that server?

gurbirkalsi:
The role for `dashboard-install` involves installing yarn (in addition to node and npm) on the remote host as part of the installation process. 

`yarn build` needs to be run locally before running the `dashboard-deploy` role and instructions for using yarn as the default dependency manager is outlined in the root `README.md`.

portante:
> `yarn build` needs to be run locally before running the `dashboard-deploy` role and instructions for using yarn as the default dependency manager is outlined in the root `README.md`.

But how does one ensure they configuration built into the UMI.js will work with the target server?  In other words, the configuration used to build UMI.js file has to work with the target it will be deployed to.

In the past, we did this by having the UMI.js file be independent of the host deployed to, and having a `config.json` file per-host deployed.

How do we ensure the same with this mechanism?

gurbirkalsi:
Previously, one would have to confirm `config.json.j2` had the correct values by testing API endpoints locally using `mock/datastoreConfig.js`. Once the user confirmed this, `config.json.j2` would be copied to the root directory of the dashboard on the target server. We maintain the same behavior by abstracting this down to a single configuration under `config/endpoints.js` which can be used for local development and in production.

The main difference between an independent config vs. integrated config is the independent config allows you to edit the configuration after a deploy to a remote host. We are required to build a new dashboard binary for every deployment to a unique target host under the integrated config approach. 

The main motivation behind the integrated config approach is to reduce overhead on page components that previously queried the config file on every component mount and allow reference of config endpoints as environment variables at any point after the node process execution. This allows us to initialize bug reporting such as Sentry at the earliest point during the node process execution and optimize the breadth of information collected. Overall, we sacrifice the ability to edit the endpoint config post deployment in order to optimize for faster component performance and earlier tools initialization.

ansible/README.md
```diff
@@ -3,20 +3,21 @@ This will ease installation, and deployment of the pbench dashboard.
 
 ## Required
 - Ansible needs to be installed on the host where you want to run this playbook
-- An inventory file containing the following key values defined:
-  - "`elasticsearch_url`", "`results_url`", "`graphql_url`", "`run_index`", "`prefix`"
-    See the `/README.md` for more details. 
+- An inventory file containing the server values defined
+
+## Endpoint Configuration
+API endpoints are configured and bundled into the final binary when running `yarn build`. Before deployment of the binary to a remote host, consider the current configuration definition at `config/endpoints.js`.
```
dbutenhof: Interesting, since I was playing with this yesterday afternoon trying to deploy a dashboard to my VM. I'd forgotten you'd merged the `process.env` change and initially built and deployed "the old way", which didn't work.

Then when I remembered and edited `endpoints.js` I found that `yarn build` didn't include the changes ... I had to delete the `dist/` directory contents and rebuild from scratch. Maybe worth a note that apparently just changing `endpoints.js` and `yarn build` isn't enough to pick up the changes?

gurbirkalsi: Running `yarn build` should overwrite the contents in `dist/` according to the UmiJS docs (`yarn build` maps to `umi build` in `package.json`). I was able to deploy the latest dashboard changes to staging without having to delete the `dist/` directory. Perhaps clearing the yarn cache might help here? https://classic.yarnpkg.com/en/docs/cli/cache/

gurbirkalsi:
@portante @dbutenhof  I've updated the PR to meet both requirements of referencing the dashboard endpoint config as a standalone component and referencing the dashboard endpoint config through environment variables in the bundled JavaScript code. 

This is accomplished by initializing a `.env` file at the root of the dashboard and using the `dotenv` library to load environment variables from `.env` into `process.env` once the node process boots up (UmiJS does not contain functionality to do this by default). The Ansible scripts have been updated to copy the `.env` file to the target server as part of the deployment process and can be modified after deployment. I've deployed the dashboard using this model to staging and if you inspect `/var/www/html/dashboard/` you should see the `.env` configuration being referenced.

.env
```diff
@@ -0,0 +1,7 @@
+# Example Configuration
+elasticsearch='http://test_domain.com'
+results='http://test_domain.com'
+graphql='http://test_domain.com'
+prefix='test_prefix.'
+result_index='test_index.'
+run_index='test_index.'
```
aquibbaig: Let's add a trailing line on this one.

dbutenhof:
> @portante @dbutenhof I've updated the PR to meet both requirements of referencing the dashboard endpoint config as a standalone component and referencing the dashboard endpoint config through environment variables in the bundled JavaScript code.
> 
> This is accomplished by initializing a `.env` file at the root of the dashboard and using the `dotenv` library to load environment variables from `.env` into `process.env` once the node process boots up (UmiJS does not contain functionality to do this by default). The Ansible scripts have been updated to copy the `.env` file to the target server as part of the deployment process and can be modified after deployment. I've deployed the dashboard using this model to staging and if you inspect `/var/www/html/dashboard/` you should see the `.env` configuration being referenced.

I like that better than embedding the parameters in the UMI build; it's much more portable. The idea that Pete and I had kicked around, which is really the basis for my [#2024](https://github.com/distributed-system-analysis/pbench/pull/2024), is to move towards a zero-(pre) configuration, ensuring the dashboard is automatically consistent with the settings of the server it's using by having the server tell it through this new API. The only parameter that would need to be statically configured in your `.env` would be the port on which the server is running so you could make that query. (E.g., just a `GET host:8001/api` or some such.) And, ideally, we'll switch over our gunicorn/apache configuration to respond to `/api...` on the default http/https port so you wouldn't even need that; you'd just do the query back to the host you loaded from.

ansible/README.md
```diff
@@ -3,20 +3,31 @@ This will ease installation, and deployment of the pbench dashboard.
...
+API endpoints are defined in `.env` as environment variables for reference in the dashboard binary. Before deployment of the binary to a remote host, consider the configuration at `.env` as the file is copied to the target server during deployment. Please reference the following example for required configuration fields:
+
+```
+elasticsearch='http://test_domain.com'
+results='http://test_domain.com'
+graphql='http://test_domain.com'
+prefix='test_prefix.'
+result_index='test_index.'
+run_index='test_index.'
+}
```
dbutenhof: `}` ?? I don't see an opening brace, so I assume this was an accident?

package.json
```diff
@@ -36,6 +36,7 @@
     "ant-design-pro": "^2.1.1",
     "antd": "^3.16.1",
     "classnames": "^2.2.5",
+    "dotenv": "^8.2.0",
```
portante: What does the `dotenv` package do?  If I look at the Network tab of a browser console, will I see GET operations fetching the `.env` file?

gurbirkalsi: `dotenv` reads the contents of the `.env` file when the Node.js process boots up. It creates an `env` object as a property of the global `process` object. This workflow doesn't involve network requests and is the reason we can reference configuration at the earliest point in dashboard initialization.

portante: But is the `.env` file built into the index.html page that is loaded?  Or the `<script ...>` references in the `index.html` back on the server?  How does Node.js running in the browser get the contents of the `.env`?  Or are you talking about a Node.js process on the server that reads this?

gurbirkalsi: The bundled dashboard code is configured to read the `.env` file at the root of the deployed dashboard directory. This is done by adding `require('dotenv').config();` to the UmiJS config which is then bundled into the `umi.js` file referenced in `index.html`. We don't require a Node.js process to run on a target server in order to serve the `.env` file.

portante: > The bundled dashboard code is configured to read the `.env` file at the root of the deployed dashboard directory. This is done by adding `require('dotenv').config();` to the UmiJS config which is then bundled into the `umi.js` file referenced in `index.html`. We don't require a Node.js process to run on a target server in order to serve the `.env` file.

So then the `umi.js` file is _built_ with the contents of the `.env` file, and then the `umi.js` file is deployed to the target server?

Do we all agree that the principle of creating a deployable `umi.js` file once and deploying it to multiple environments is one we want to adhere to as a project?

If we all agree, great.

If we don't agree with that principle, we need to discuss that first because we need a guide for how we handle builds and deployment.

That principle is at the heart of containers (build the container once, deploy it via different orchestrators [podman, kubernetes, others]), of Ansible (roles and playbooks are applied to various inventory files with different contents), and CI/CD pipelines (build once, test the same object without changes through the pipeline until deployed).

dbutenhof: > > The bundled dashboard code is configured to read the `.env` file at the root of the deployed dashboard directory. This is done by adding `require('dotenv').config();` to the UmiJS config which is then bundled into the `umi.js` file referenced in `index.html`. We don't require a Node.js process to run on a target server in order to serve the `.env` file.
> 
> So then the `umi.js` file is _built_ with the contents of the `.env` file, and then the `umi.js` file is deployed to the target server?

Hmm; good question, but there is a `/var/www/html/dashboard/.env` on the staging server, and my impression was that `dotenv` is somehow sucking that in. I don't really see how that could happen with a browser and the static dashboard Javascript on the server unless it's doing an HTML GET to retrieve it just like it does the CSS.

> Do we all agree that the principle of creating a deployable `umi.js` file once and deploying it to multiple environments is one we want to adhere to as a project?

Yes, very strongly!

gurbirkalsi: I agree with the "build once, deploy anywhere" paradigm. The aim of this PR is to bridge the gap between deploying to multiple environments with one binary while also allowing endpoint config to be referenced at dashboard initialization. The discussion in this issue captures why this is an uncommon pattern in React environments and generally requires a "hacked" workaround: https://github.com/facebook/create-react-app/issues/2353

I've reviewed suggested workarounds and come up with a solution to configure endpoints at dashboard runtime so we avoid bundling endpoints into the dashboard binary. We do this by referencing ` <script src="<%= context.config.publicPath +'endpoints.js'%>"></script>` in the dashboard `index.html` template and fetching `endpoints.js` with a `GET` request. `endpoints.js` is included in `dist/` as a standalone file and can be modified after deployment.

I've deployed the dashboard to staging using this model. Feel free to inspect the Network tab in dev tools to see the `endpoints.js` file being queried. Additionally, you can modify `endpoints.js` and refresh the dashboard to try different API endpoints.

portante (approved):

aquibbaig (approved):

src/pages/document.ejs
```diff
@@ -6,7 +6,10 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Pbench Dashboard</title>
-  <link rel="icon" href="/favicon.ico" type="image/x-icon">
+  <link rel="icon" href="<%= context.config.publicPath +'favicon.ico'%>" type="image/x-icon">
+  
+  <!-- runtime endpoints config -->
+  <script src="<%= context.config.publicPath +'endpoints.js'%>"></script>
```
dbutenhof, line 12: I haven't seen that `<%=` syntax before. What does that actually mean? I tried looking it up, but what I found wasn't really helpful. Is it just a Javascript environment to allow the path concatenation operator?

dbutenhof (approved):

src/pages/document.ejs
gurbirkalsi: This has been adapted from a framework called [ejs](https://github.com/mde/ejs). It allows for dynamic HTML templating and has some pretty cool features!

PR #111, opened 2020-12-09
Visualize ToC data with a react-based tree visualizer 

anishaswain:
- [ ] Fetching child data on click of the parent result

- [ ] Combine child and parent data to create a new tree on click on the parent result

- [ ] Searching the TOC tab with keywords entered by the user.

- [ ] Filtering the TOC data with file extension from the dropdown.

- [ ] Link items in the TOC tab to their file on the results server

- [ ] Any collapsed item should also collapse all of its children nodes in the tree view


pravins:
This is fixed with https://github.com/distributed-system-analysis/pbench/pull/2952
We have a new design now.
We are opening new bug/rfe   Filtering the TOC data with file extension from the dropdown.

PR #114, opened 2020-12-18
Dashboard support for ES7 / Pbench server 0.71

### LINKED ISSUE #71: Upgrade Elasticsearch to v7
Update queries to Elasticsearch with v7 features

dbutenhof:
Revise dashboard Elasticsearch queries to be compatible with Elasticsearch 7 syntax in order to work with 0.71 Pbench server, which now indexes to Elasticsearch 7.

Of special note: this also switches the dashboard over to use the two Pbench server "native query" APIs I've implemented, replacing both the `query` services and the `fetch` model effect transformations for `queryControllers`/`fetchControllers` and `queryMonthIndices`/`fetchMonthIndices`.

Also, during testing I found inconsistencies in the search behavior that seem to have been pre-existing bugs; e.g., the search state passed on to `fetchIterationSamples` didn't contain the `id` property required by that code to construct the queries. (Pay special attention to these changes.)

The dashboard actually runs on my 0.71 server VM and the RDU2 elasticsearch instance. The `yarn test` passes, with a warning for the `SiderMenu` test which apparently isn't including a `key` prop in a component list: not my change, and I decided not to mess with it. The e2e test script passes locally; hopefully will also on TravisCI.

The new `endpoints.js` should look something like this:

```
window.endpoints = {
  elasticsearch: 'http://elasticsearch.intlab.perf-infra.lab.eng.rdu2.redhat.com',
  results: 'http://dhcp31-170.perf.lab.eng.bos.redhat.com:8901',
  graphql: 'http://graphql.perf.lab.eng.bos.redhat.com:9466',
  prefix: 'drb.',
  result_index: 'v5.result-data-sample.',
  result_data_index: 'v5.result-data.',
  run_index: 'v6.run-data.',
  run_toc_index: 'v6.run-toc.',
  pbench_server: 'http://dhcp31-170.perf.lab.eng.bos.redhat.com:8001/api/v1',
};
```

Note that I changed the placeholders for the index strings to be a bit more useful. Ideally, we'll be able to convert the dashboard over to a server-driven configuration like my [#2024](https://github.com/distributed-system-analysis/pbench/pull/2024) endpoint configuration API, which would remove all of this complexity.

Also of note is my [Server API Architecture](https://docs.google.com/document/d/1c0bauJ9DM7MckFKus8XuTfkqZuLvVONnpnRDoM_TCik/edit#) document, which has before/after payload samples and additional notes.

### DIFF
--- a/public/endpoints.js
+++ b/public/endpoints.js
@@ -3,6 +3,9 @@ window.endpoints = {
   results: 'http://test_domain.com',
   graphql: 'http://test_domain.com',
   prefix: 'test_prefix.',
-  result_index: 'test_index.',
-  run_index: 'test_index.',
+  run_index: 'vn.run-data.',
+  run_toc_index: 'vn.run-toc.',
+  result_index: 'vn.result-data-sample.',
+  result_data_index: 'vn.result-data.',
+  pbench_server: 'http://test_domain.com/api/v1',
 };
--- a/src/models/search.js
+++ b/src/models/search.js
@@ -23,14 +23,14 @@ export default {
       const { endpoints } = window;
 
       const index = endpoints.prefix + endpoints.run_index + indices[0];
-      const mapping = response[index].mappings['pbench-run'].properties;
-      const fields = [];
+      const mapping = response[index].mappings.properties;
+      let fields = [];
       const filters = {};
 
       Object.entries(mapping).forEach(([key, value]) => {
         if (typeof value.properties !== 'undefined') {
           filters[key] = Object.keys(value.properties);
-          fields.concat(Object.keys(value.properties));
+          fields = fields.concat(Object.keys(value.properties));
         }
       });
 
@@ -52,21 +52,40 @@ export default {
       const { selectedFields } = payload;
 
       const searchResults = {};
-      searchResults.resultCount = response.hits.total;
+      searchResults.resultCount = response.hits.total.value;
       const parsedResults = [];
+
+      const getValue = (data, keys) => {
+        let value;
+        if (typeof data[keys[0]] === 'undefined') {
+          value = undefined;
+        } else if (keys.length === 1) {
+          value = data[keys[0]];
+        } else {
+          value = getValue(data[keys[0]], keys.slice(1));
+        }
+        return value;
+      };
 
       response.hits.hits.forEach(result => {
         const parsedResult = {};
-
         selectedFields.forEach(field => {
-          if (typeof result.fields[field] !== 'undefined') {
-            const fieldValue = result.fields[field][0];
-            parsedResult[field] = fieldValue;
+          const value = getValue(result._source, field.split('.'));
+          if (typeof value !== 'undefined') {
+            parsedResult[field] = value;
           }
         });
 
+        /*
+         * NOTE: these make the output of searchQuery compatible with the
+         * expectations of fetchIterationSamples, as it would normally get
+         * from fetchResults
+         */
         if (typeof result._id !== 'undefined') {
-          parsedResult.key = result._id;
+          parsedResult.id = result._id;
+        }
+        if (typeof result._source.run.name !== 'undefined') {
+          parsedResult.key = result._source.run.name;
         }
 
         parsedResults.push(parsedResult);
--- a/src/services/dashboard.js
+++ b/src/services/dashboard.js
@@ -10,7 +10,7 @@ function scrollUntilEmpty(data) {
   const endpoint = `${endpoints.elasticsearch}/_search/scroll?scroll=1m`;
   const allData = data;
 
-  if (allData.hits.total !== allData.hits.hits.length) {
+  if (allData.hits.total.value !== allData.hits.hits.length && allData._scroll_id) {
     const scroll = request.post(`${endpoint}&scroll_id=${allData._scroll_id}`);
     scroll.then(response => {
       allData._scroll_id = response._scroll_id;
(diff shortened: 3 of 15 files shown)


src/services/dashboard.js
```diff
@@ -206,7 +163,7 @@ export async function queryIterationSamples(params) {
                   aggs: {
                     title: {
                       terms: {
-                        field: 'sample.measurement_title',
+                        field: 'sample.measurement_title_kw',
```
portante: Base on https://github.com/distributed-system-analysis/pbench/pull/2053, shouldn't this be `'sample.measurement_title.measurement_title_kw'`?

portante (changes requested):

src/services/dashboard.js
dbutenhof: Oddly, ES7 didn't complain about this query, so maybe it really does work as I thought I'd seen last year. However, as you'd pointed out in the other PR, current ES7 documentation references it as a scoped sub-field name, so I'm changing this. I thought the name was pretty awkward, anyway. ;-)

dbutenhof:
So Travis failed at `yarn install` ... this is clearly a Travis environmental difference that doesn't appear to have anything to do with my changes. Any ideas?

```
warning file-loader@2.0.0: Invalid bin field for "file-loader".
warning umi-url-pnp-loader@1.1.2: Invalid bin field for "umi-url-pnp-loader".
info fsevents@2.3.1: The platform "linux" is incompatible with this module.
info "fsevents@2.3.1" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@1.2.13: The platform "linux" is incompatible with this module.
info "fsevents@1.2.13" is an optional dependency and failed compatibility check. Excluding it from installation.
error postcss@8.2.2: The engine "node" is incompatible with this module. Expected version "^10 || ^12 || >=14". Got "13.14.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
yarn install failed!
```

src/models/search.js
```diff
@@ -23,16 +23,18 @@ export default {
       const { endpoints } = window;
 
       const index = endpoints.prefix + endpoints.run_index + indices[0];
-      const mapping = response[index].mappings['pbench-run'].properties;
-      const fields = [];
+      console.log(`index: ${index}, mapping: ${JSON.stringify(response, null, 2)}`);
```
portante: Is this debugging only?

public/endpoints.js
```diff
@@ -3,6 +3,9 @@ window.endpoints = {
   results: 'http://test_domain.com',
   graphql: 'http://test_domain.com',
   prefix: 'test_prefix.',
-  result_index: 'test_index.',
-  run_index: 'test_index.',
+  run_index: 'vn.run-data.',
+  run_toc_index: 'vn.run-toc.',
+  result_index: 'vn.result-data-sample.',
+  result_data_index: 'vn.result-data.',
```
portante, line 9: These five above don't seem like end points.  This seems like what we want the initial server load to give the dashboard.

portante (changes requested):

src/models/search.js
dbutenhof: I knew I forgot something!

public/endpoints.js
dbutenhof: Yeah, "endpoints" might be better termed "server configuration". More obvious now that I split the two main indices. In my configuration API I put these in a "metadata" section separate from the "endpoints" (URI) section.

aquibbaig: Hey @dbutenhof, I suppose this file is in `.gitignore`

dbutenhof: > Hey @dbutenhof, I suppose this file is in `.gitignore`

Yes, it is; although for some reason that pattern isn't working for me. Which in this case is fine as I had to add additional configuration keys.

portante: Wait, if public/endpoints.js is in .gitignore, why are we not removing it instead of changing it in the PR?

portante (changes requested):

public/endpoints.js
dbutenhof: Actually, that's a good point; .gitignore is only for _untracked_ files. Maybe we should really have a `public/endpoints.example.js` that's tracked and updated, with the build looking for an untracked `public/endpoints.js` that's in .gitignore.

dbutenhof: Hmm; that, unfortunately, breaks the current end-to-end testing because the search test requires `endpoints.js` placeholders that match the internal test framework placeholders. (This is actually a problem if you run the tests with a "real" `endpoints.js`, too.) It's probably possible to get npm/yarn/umi to "build" the example file as the `endpoints.js` to use, but I don't know enough about the capabilities and structure of `package.json` to do that.

gurbirkalsi: I'll take a look into possibly using a mocked endpoint file for enabling e2e testing without the requirement of `endpoints.js`. @dbutenhof we should update the fields present in the README.md file at the root of the project directory.

src/models/search.js
```diff
@@ -55,18 +55,37 @@ export default {
       searchResults.resultCount = response.hits.total;
```
gurbirkalsi: It seems `resultCount` is being assigned an object opposed to an integer specified from the `total` field. This is causing `[Object] [object]` to be rendered in the Search page component once a search query has been made.

gurbirkalsi (changes requested):
Travis CI is failing due to an incompatible version of node running the test environment.

```
error postcss@8.2.3: The engine "node" is incompatible with this module. Expected version "^10 || ^12 || >=14". Got "13.14.0"
```

We should consider upgrading the node environment to >=14 in `.travis.yml` to resolve this issue.

src/models/search.js
dbutenhof: Thanks. Somehow I missed that, both visually and comparing my Postman ES7 output... hits.total is an object `{value: 1, relation: 'eq'}` ... I want `response.hits.total.value` here.

dbutenhof: image

gurbirkalsi (approved):
LGTM, just needs a rebase to resolve TravisCI issues!

dbutenhof:
Well, this is interesting. I had done some brief experimentation with renaming `endpoints.js` to work with the intent of the `.gitignore`. And apparently when I pushed my rebase, I ended up actually deleting it... not just from github but from my local repo. Which, of course, breaks the e2e tests.

So now I've got to restore it to match the internal mocking. It'll be a minute...

dbutenhof:
I suppose it's worth a note here that when this PR is merged, the dashboard `master` will no longer work with Pbench server code earlier than `master` ... the current production and staging servers are 0.69 and run with Elasticsearch V1.

Should we cut a dashboard branch before merging this? And possibly even updating it to bump the dashboard version?

dbutenhof:
> I suppose it's worth a note here that when this PR is merged, the dashboard `master` will no longer work with Pbench server code earlier than `master` ... the current production and staging servers are 0.69 and run with Elasticsearch V1.
> 
> Should we cut a dashboard branch before merging this? And possibly even updating it to bump the dashboard version?

FYI: After consulting with Gurbir, I've _updated the dashboard version to 3.0.0_ for this change.

Issue #115, opened 2021-01-11
Add localization feature to Pbench Dashboard


anishaswain:
Internationalization or Localization will make the Pbench Dashboard reachable to people understanding different languages.

anishaswain:
@gurbirkalsi  @portante here are the few libraries I found to help us achieve internationalization with React
[react-i18next](https://www.npmjs.com/package/react-i18next), [format.js](https://github.com/formatjs/formatjs), [react-localization](https://www.npmjs.com/package/react-localization)

PR #116, opened 2021-01-11
Fixes login page 

anishaswain:
_Fixes:_
1. Login options button with padding and margin
2. Adds a smaller font for "forgot password"
3. Comments out dropdown for language selection

Pbench-Dashboard (1)

### DIFF
--- a/src/components/AuthLayout/index.js
+++ b/src/components/AuthLayout/index.js
@@ -45,21 +45,6 @@ class AuthLayout extends Component {
   render() {
     let { heading } = this.props;
     const { backOpt, toPreview } = this.props;
-
-    // P.S. : Will be added later with drpdown for language localization
-    // const { isOpen } = this.state;
-    // const dropdownItems = [
-    //   <DropdownItem key="german" className={styles.form}>
-    //     German
-    //   </DropdownItem>,
-    //   <DropdownItem key="hindi" className={styles.form}>
-    //     Hindi
-    //   </DropdownItem>,
-    //   <DropdownItem key="french" className={styles.form}>
-    //     French
-    //   </DropdownItem>,
-    // ];
-
     const back = (
       <Button
         id="backBtn"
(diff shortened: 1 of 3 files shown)


gurbirkalsi (approved):

aquibbaig (approved):
Looks good!

src/components/AuthLayout/index.js
```diff
@@ -142,11 +134,7 @@ class AuthLayout extends Component {
         </GridItem>
         <GridItem>
           <Title headingLevel="h4" size="xl">
-            <Button
-              variant="link"
-              className={styles.inlineLink}
-              onClick={() => this.navigate('password')}
-            >
+            <Button variant="link" onClick={() => this.navigate('password')}>
```
dbutenhof, line 103: So it's smaller just because we're _not_ giving it the `inlineLink` style? Would it be better if the size was explicit rather than dependent on the surrounding context?

dbutenhof (approved):

gurbirkalsi:
@anishaswain, seems we need a rebase in order to resolve the Travis CI build errors.

src/components/AuthLayout/index.js
```diff
@@ -88,7 +78,9 @@ class AuthLayout extends Component {
           {backOpt === 'true' ? back : <Fragment />}
           <br />
           {heading === undefined ? (heading = 'login with...') : heading}
-          <Dropdown
+
```
portante: Not convinced it is worth keeping code that is commented out.  Perhaps file an issue to track what you'd need to do later for localization?

src/components/AuthLayout/index.js
```diff
@@ -56,19 +45,20 @@ class AuthLayout extends Component {
...
-        German
-      </DropdownItem>,
-      <DropdownItem key="hindi" className={styles.form}>
-        Hindi
-      </DropdownItem>,
-      <DropdownItem key="french" className={styles.form}>
-        French
-      </DropdownItem>,
-    ];
+    // P.S. : Will be added later with drpdown for language localization
```
portante: Not sure about keeping commented-out code ... (see below).

portante (changes requested):

src/components/AuthLayout/index.js
anishaswain: removed and will be tracked with #115

src/components/AuthLayout/index.js
anishaswain: removed and will be tracked with #115

src/components/AuthLayout/index.js
aquibbaig: I think if we are not using that, we could consider removing the `inlineLink` css definition

aquibbaig:
@anishaswain I tried to reproduce the unit test issue and I think the Dropdown component is missing from the component due to which the unit test is failing

PR #117, opened 2021-01-14
Update createdAt and updatedAt fields with defaults for Sessions type

gurbirkalsi:
Converts session sharing components to use PatternFly React


aquibbaig (approved):
Looks great! :tada:

portante (approved):
This looks good as is, but I am wondering if we should consider a name change away from "sessions" here, and using a name that is tied closer to the nature of the data being reviewed?

gurbirkalsi:
I'm certainly open to any naming suggestions here. Since a session URL can be created at any point within the dashboard 'state', it becomes difficult to tie naming of the field to a specific workflow as there are many (run comparison, search, summary, etc).

PR #121, opened 2021-02-01
enhancements for the signup page

aquibbaig:
This PR fixes the following issues/makes the following enhancements:
- Once the user types the "password" and "confirmPassword" fields, if he again changes the password field for some reason, the validation doesn't re-render, saying "passwords do not match".
- Prevents form from multiple re-renders due to a lot of variables for each input change.  (Errors and constraints now are managed as one object only)
- Converted the whole component into a function component and uses `mapStateToProps` to connect to the store.
- References the `lastName` field, which was missing before

I have disabled `eslint` for the form validation which complained against using a `for...in` loop, however considering the fact that we only have 4 Object.entries(), that loop should not be a nuisance.

### DIFF
--- a/src/pages/SignupHandler/index.js
+++ b/src/pages/SignupHandler/index.js
@@ -1,4 +1,7 @@
-import React from 'react';
+import React, { useState, useEffect, Fragment } from 'react';
+import AuthLayout from '@/components/AuthLayout';
+import { CheckIcon, CloseIcon } from '@patternfly/react-icons';
+import { connect } from 'dva';
 import {
   Form,
   FormGroup,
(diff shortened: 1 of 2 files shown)


src/pages/SignupHandler/index.js
```diff
@@ -9,265 +12,295 @@ import {
...
+      }
+    }
+    // check if all constraints are met.
+    for (const ct of Object.entries(constraints)) {
+      if (ct[1] !== 'met') {
+        return false;
+      }
+    }
+    // if we reach here, it means
+    // we have convered all of the edge cases.
```
dbutenhof: "convered" => "covered"

src/pages/SignupHandler/index.js
```diff
@@ -9,265 +12,295 @@ import {
...
+        <TextInput
+          isRequired
+          type="password"
+          id="horizontal-form-confirm-password"
+          name="horizontal-form-confirm-password"
+          onChange={checkConfirmPassword}
+        />
+        <p className={styles.error}>{errors.passwordConfirm}</p>
+      </FormGroup>
+      <div id="passworkCheck">
```
dbutenhof: This is just reorganized a bit from before, and you didn't change the line; but this should probably be "passwordCheck"; although I don't know where else the `id` might be referenced.

src/pages/SignupHandler/index.js
```diff
@@ -9,265 +12,295 @@ import {
...
+  );
+
+  const getSplitDisplayProperty = (name, type) => {
+    if (constraints[name] === 'met' && type === 'check') {
+      return 'block';
+    }
+    if (constraints[name] === 'unmet' && type === 'close') {
+      return 'block';
+    }
+    return 'none';
```
dbutenhof, line 96: The key to understanding this was the interpretation of the `display` CSS property; and while I've seen it before, I don't use it a lot and I had to go look up the distinction of `block` (display as a block) vs `none` (don't display at all) and then work that into the logic around the green and red icons for "yeah we're good" or "you missed this constraint".

A comment on the method could make this a lot easier for readers who also aren't that familiar with it.

dbutenhof (approved):

src/pages/SignupHandler/index.js
aquibbaig: Right. I would change these typos. And dereference some unused ids

src/pages/SignupHandler/index.js
aquibbaig: Sure, thanks for bringing this up.

src/pages/SignupHandler/index.js
aquibbaig: I smh always end up doing typos

src/pages/SignupHandler/index.js
```diff
@@ -9,265 +12,295 @@ import {
...
+  const handleLastNameInputChange = val => {
+    setLastName(val);
   };
 
-  handleEmailInputChange = email => {
-    const { username, password, confPassword } = this.state;
-    this.setState({
-      email,
-    });
+  const handleEmailInputChange = val => {
```
gurbirkalsi, line 107: It seems there's an edge case in this section causing the registration form to proceed when inputing a correct email and then removing it. Can we debug this further? 

```
Steps to reproduce:
1. type "testemail@redhat.com" in email address input field
2. fill all other fields correctly
3. remove text from email address input field
4. click "Create account" button
```

gurbirkalsi (changes requested):

src/pages/SignupHandler/index.js
aquibbaig: Whoa, Thanks Gurbir. It was in the back of my head but I smh forgot about it in the process.
This should fix it.
```diff
     () => {
       if (validateForm()) {
         setEnableSubmitButton(true);
+      } else {
+        setEnableSubmitButton(false);
       }
     },
     [firstName, lastName, email, password, confirmPassword]

```

dbutenhof (approved):

gurbirkalsi (approved):

PR #123, opened 2021-02-04
add password recovery boilerplate

### LINKED ISSUE #122: Forgot password mechanism on the frontend
We would certainly need an implementation of "forgot password" workflow on the frontend and let our dashboard handle that. Some of these steps include the following:
- When the user clicks on forgot password option in the login page, he is taken to a route where he is prompted for username/email or both based on the validation that we will do on the backend.
- Backend generates a link which takes the user to the password recovery page for the actual account. (It is important to validate the authenticity of the link which is generated in the backend somehow).
- After the user recovers his password successfully, he will be taken to the login page, where he would log in with his new credentials

aquibbaig:
Initial PR for fixing #122


src/pages/PasswordRecovery/index.js
```diff
@@ -0,0 +1,247 @@
...
+        <FormGroup label="Password" isRequired fieldId="horizontal-form-password">
+          <TextInput isRequired type="password" onChange={handlePasswordChange} />
+          <p className={styles.error}>{errors.password}</p>
+        </FormGroup>
+        <FormGroup label="Confirm Password" isRequired fieldId="horizontal-form-confirm-password">
+          <TextInput isRequired type="password" onChange={handleConfirmPasswordChange} />
+          <p className={styles.error}>{errors.confirmPassword}</p>
+        </FormGroup>
+        <div>
+          <h4>Password must contain at least</h4>
```
portante: Any way we can leverage the same from the normal signup page?

portante (changes requested):

src/pages/PasswordRecovery/index.js
aquibbaig: Do we create this (split components for password constraints) as a separate component and reference them in each page component that we need?

aquibbaig: Fixed!

config/router.config.js
```diff
@@ -15,9 +15,14 @@ module.exports = [
...
   {
-    path: '/password',
-    name: 'password',
-    component: './PasswordHandler',
+    path: '/recovery_email',
+    name: 'recovery_email',
+    component: './RecoveryEmail',
+  },
+  {
+    path: '/recovery/:hash',
```
aquibbaig: should be `reset_password/:hash`

webbnh (dismissed):
As far as I can tell (which might not be that far...) it looks fine.  The password change code could be slightly DRYer, but I'm not inclined to worry about that.

src/services/auth.js
```diff
@@ -45,3 +45,20 @@ export const postLogoutUser = () => {
...
   });
 };
+
+export const putUpdateNewPassword = params => {
+  const { temporaryToken, password } = params;
+  const endpoint = `${endpoints.api.reset_password}`;
+  return request.put(endpoint, {
+    data: {
+      // First name is expected
+      // on the server side.
```
webbnh: I'm not sure what this code comment is supposed to be telling me.  Aside from the fact that I believe that the Server expects the _user_ name (not the "First name"), if the server expects it why isn't the call supplying it?

webbnh (dismissed):
Just a question....

src/services/auth.js
webbnh: It appears that this is being tracked in #211.  Until that's resolved, it would probably be good to reference it in this code comment.

webbnh (approved):
👍

pravins:
We are moving to Keycloak, so no need for this PR.

PR #124, opened 2021-02-04
react-table integration with PatternFly CSS 

gurbirkalsi:
This PR introduces the `react-table` library to handle table visualizations across the dashboard. In order to provide the PatternFly look and feel, PatternFly CSS from `@patternfly/patternfly` is used to apply CSS to various HTML elements (`<table>, <tbody>, <tr>, etc.`). 

e2e test selectors have been updated with new references to `react-table` in order to resolve current issues and all current open PRs will require a rebase in order to reference the new e2e environment.  

Once this component is merged with the rest of the dashboard, we will be able to remove `ant-design` from our dependency tree along with any remaining references to Ant Design components within the dashboard.


portante (commented):
Great start; @anishaswain, it would be great to work with this with the ToC code, too.

portante (commented):
Is this ready be to tried out with the ToC work from @anishaswain?

gurbirkalsi:
We need to integrate the `expandable` plugin from `@patternfly/react-table` in order to make table rows expandable for displaying nested children contents. I'll take a look at integrating this with the Table component next.

portante (commented):
Do you have any screen shots on how this might look and behave?

anishaswain (approved):
The PR looks great @gurbirkalsi. However I do feel that the design is a bit off and needs CSS fixes to make it look more like patternfly table
1. The tables have lot of white spaces. May be adding borders will help resolve that
Screenshot 2021-04-08 at 1 44 58 PM
2. In some places the headers are cropped. We need to wrap the texts to acomodate themselves according to the width of the column

Pbench Dashboard


However we can address the CSS changes in another PR.

portante (approved):

src/components/Table/index.js
```diff
@@ -1,29 +1,245 @@
...
-      loading={loading}
-      onRow={onRow}
-      scroll={{
-        x: true,
-      }}
-      pagination={{
-        defaultPageSize: 10,
-        showSizeChanger: true,
-        pageSizeOptions: constants.tableSizeOptions,
+    <SearchBar
```
aquibbaig, line 31: I personally liked the previous design with the SearchBar being on the same row as "DatePicker". I understand why we moved away from that (as SearchBar is a part of the Table component), I believe at least we should apply a "margin-top" property to separate it from the "DatePicker component up top". 

React table

src/components/Table/index.less
```diff
@@ -0,0 +1,16 @@
+tr:hover td {
+  font-weight: bold;
+}
```
aquibbaig: Adding this property on hover causes the whole table to shift laterally some pixels, I believe changing the row color on focus should be enough visual cue and we can drop this change.

aquibbaig: Steps to reproduce this:
In the controllers tab, hover on the name of the controller as it turns bold, sometimes when the name is too large, it causes the table to shift in horizontal direction by some amount of pixels.

Additional information:
I inspected the source maps to find where this css was coming from, to my surprise it was referenced by `index.css` loaded by `localhost:8000/node_modules/jschart`. (no idea why)

aquibbaig (changes requested):
LGTM @gurbirkalsi! This is great. Just some minor issues which we can fix here. 

- I am not able to favorite records as of now.
- When I click on any controller, it takes me to the results page with the heading `[Object object]`.
- Also, can you please rebase this branch on top of `main` to reference correct config variables?

src/components/Table/index.js
gurbirkalsi: Thanks for catching this @aquibbaig, it seems we had a typo when passing style props as `styles` whereas we needed `style`.

src/components/Table/index.less
gurbirkalsi: I've updated the Table to now handle hovering using the `"pf-m-hoverable"` class name. This should allow us to maintain the appropriate visual key for hovering and highlighting a controller according to PatternFly best practices.

gurbirkalsi:
@anishaswain This PR is only using PatternFly CSS dependencies and only injects custom CSS for handling overflows when the number of table rows exceed the page height. Here is the CSS & HTML being referenced from PatternFly's documentation: https://www.patternfly.org/v4/components/table/html#

In the case we need more visual cues or adjustments to design, I'd recommend filing an issue at https://github.com/patternfly/patternfly-design to address white spaces.

gurbirkalsi:
@aquibbaig Unfortunately, I wasn't able to reproduce the favoriting records issue and results page with the heading `[Object object]`. Were there any logging errors or output that might be helpful in debugging this?

aquibbaig (approved):
LGTM! @gurbirkalsi

aquibbaig:
> @aquibbaig Unfortunately, I wasn't able to reproduce the favoriting records issue and results page with the heading `[Object object]`. Were there any logging errors or output that might be helpful in debugging this?

This issue has been resolved now. Just tested it out

PR #126, opened 2021-02-06
User authentication in the frontend

aquibbaig:
This PR registers a user to Pbench from the dashboard.

TODOs:
- [ ] Discuss for the naming convention of `usernames` that we must give to the users. Currently, it is `firstName.toLowerCase()`
- [ ] Snack bars to display visual cues to users (if successfully registered or some failure occurred).
- [x] Do not throw an exception on errors while registering, rather display the error message and handle the error gracefully. (Needs Discussion)
- [x] Need to replace the endpoints from 127.0.0.1

### DIFF
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,10 +1,11 @@
-import queryRegisterUser from '../services/auth';
+import { queryRegisterUser, queryLoginUser } from '../services/auth';
 
 export default {
   namespace: 'auth',
 
   state: {
     auth: { username: '' },
+    token: '',
   },
 
   effects: {
@@ -13,6 +14,11 @@ export default {
         type: 'modifyUser',
         payload,
       });
+    },
+    *loginUser({ payload }, { call }) {
+      const response = yield call(queryLoginUser, payload);
+      localStorage.setItem('token', response.auth_token);
+      return response;
     },
     *registerUser({ payload }, { call }) {
       const response = yield call(queryRegisterUser, payload);
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,11 +1,10 @@
-import { queryRegisterUser, queryLoginUser } from '../services/auth';
+import { queryRegisterUser, queryLoginUser, queryLogoutUser } from '../services/auth';
 
 export default {
   namespace: 'auth',
 
   state: {
     auth: { username: '' },
-    token: '',
   },
 
   effects: {
@@ -24,9 +23,16 @@ export default {
       const response = yield call(queryRegisterUser, payload);
       return response;
     },
-    *logoutUser({ payload }, { put }) {
+    *logoutUser({ payload }, { call }) {
+      const response = yield call(queryLogoutUser, payload);
+      // TODO: Must be removed from the backend through
+      // the response header.
+      localStorage.removeItem('token');
+      return response;
+    },
+    *removeUserFromStore({ payload }, { put }) {
       yield put({
-        type: 'removeUser',
+        type: 'modifyUser',
         payload,
       });
     },
@@ -37,11 +43,6 @@ export default {
       return {
         ...state,
         auth: payload,
-      };
-    },
-    removeUser() {
-      return {
-        auth: { username: '' },
       };
     },
   },
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,3 +1,5 @@
+import queryRegisterUser from '../services/auth';
+
 export default {
   namespace: 'auth',
 
@@ -11,6 +13,10 @@ export default {
         type: 'modifyUser',
         payload,
       });
+    },
+    *registerUser({ payload }, { call }) {
+      const response = yield call(queryRegisterUser, payload);
+      return response;
     },
     *logoutUser({ payload }, { put }) {
       yield put({
--- a/src/pages/LoginHandler/index.js
+++ b/src/pages/LoginHandler/index.js
@@ -37,8 +37,9 @@ class LoginHandler extends React.Component {
     dispatch({
       type: 'auth/loadUser',
       payload,
+    }).then(() => {
+      dispatch(routerRedux.push(`/private`));
     });
-    dispatch(routerRedux.push(`/private`));
   };
 
   enableSubmitBtn = () => {
@@ -83,11 +84,19 @@ class LoginHandler extends React.Component {
 
   handleLoginSubmit = () => {
     const { username, password } = this.state;
-    if (username === 'admin' && password === 'admin') {
-      this.setLoggedIn({ username });
-    } else {
-      alert('Wrong username/password pair');
-    }
+    const { dispatch } = this.props;
+    dispatch({
+      type: 'auth/loginUser',
+      payload: {
+        username,
+        password,
+      },
+    }).then(response => {
+      const { status } = response;
+      if (status === 'success') {
+        this.setLoggedIn({ username });
+      }
+    });
   };
 
   render() {
--- a/src/pages/LoginHandler/index.js
+++ b/src/pages/LoginHandler/index.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import {
   Form,
   FormGroup,
@@ -7,33 +7,31 @@ import {
   ActionGroup,
   Button,
   Title,
+  Alert,
+  AlertActionCloseButton,
 } from '@patternfly/react-core';
 import AuthLayout from '@/components/AuthLayout';
 import { routerRedux } from 'dva/router';
 import { connect } from 'dva';
 import styles from './index.less';
 
-@connect(auth => ({
-  auth: auth.auth,
-}))
-class LoginHandler extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      username: '',
-      password: '',
-      variantVal: 'tertiary',
-      btnColor: 'black',
-      btnDisabled: 'true',
-    };
-  }
+const mapStateToProps = state => {
+  const { auth } = state;
+  return auth;
+};
 
-  componentDidMount = () => {
-    this.disableSubmitBtn();
-  };
+const LoginHandler = props => {
+  const [username, setUsername] = useState('');
+  const [password, setPassword] = useState('');
+  const [btnDisabled, setBtnDisabled] = useState(true);
+  const [alertOpen, setAlertOpen] = useState(false);
+  const [alertComponent, setAlertComponent] = useState({
+    message: '',
+    variant: '',
+  });
 
-  setLoggedIn = payload => {
-    const { dispatch } = this.props;
+  const setLoggedIn = payload => {
+    const { dispatch } = props;
     dispatch({
       type: 'auth/loadUser',
       payload,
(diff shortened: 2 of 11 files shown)


src/pages/SignupHandler/index.js
```diff
@@ -145,7 +145,21 @@ const SignupHandler = props => {
...
+      type: 'auth/registerUser',
+      payload: {
+        firstName,
+        lastName,
+        username: firstName.toLowerCase(),
+        password,
+        email,
+      },
+    }).then(resp => {
+      if (resp.message === 'Successfully registered.') {
```
aquibbaig: Should ideally display a message on the dashboard as "Signed in successfully", "failed to sign in" otherwise. Should also ensure that we do not reach the exceptions page by any means

src/utils/request.js
```diff
@@ -51,4 +51,26 @@ const request = extend({
   errorHandler, // extend default error handler with custom actions
 });
 
+request.interceptors.request.use((url, options) => {
+  const token = localStorage.getItem('token');
+  if (token) {
+    try {
+      const tokenHeader = {
+        token,
```
dbutenhof: shouldn't this be `'Authorization: Bearer ' + token`? Or is there some additional magic happening behind the scenes?

src/utils/request.js
```diff
@@ -51,4 +51,26 @@ const request = extend({
   errorHandler, // extend default error handler with custom actions
 });
 
+request.interceptors.request.use((url, options) => {
+  const token = localStorage.getItem('token');
```
portante, line 78: As we discussed, this token should probably not be stored in local storage in the browser, just in Javascript memory via a variable.

portante (changes requested):

src/utils/request.js
aquibbaig: Yes, this won't be stored here. For now I've stored it here for testing purposes. When we are ready, this is gonna be a js variable in memory.

src/utils/request.js
aquibbaig: Yes Dave, this is just for testing purposes, ideally this will be a Bearer + tokenValue

aquibbaig: Fixed now!

src/services/dashboard.js
```diff
@@ -24,17 +24,14 @@ function scrollUntilEmpty(data) {
 
 export async function queryControllers(params) {
   try {
-    const { selectedDateRange } = params;
+    const { selectedDateRange, username } = params;
 
     return request.post(`${endpoints.pbench_server}/controllers/list`, {
       data: {
-        user: 'username', // TODO: Will need to get user context here
+        user: username || 'public',
```
portante: This is hardcoding a user name that might change.  We should let the server side API code handle any defaults.

portante (changes requested):
We need to fix the token storage mechanism and remove the `public` user default.

src/services/dashboard.js
dbutenhof: Yes, that's a good point; we can default a missing `username` in the payload to `public` and simply omit it here or give a null value.

aquibbaig: +1

src/models/auth.js
```diff
@@ -28,6 +43,7 @@ export default {
       };
     },
     removeUser() {
+      localStorage.removeItem('token');
```
dbutenhof: Interesting that `logoutUser` uses a reducer to remove the token from `localStorage` but `loginUser` doesn't use one to store the token. These should be consistent.

I really noticed this inconsistency because the reducer name is potentially confusing: it looks like we're taking some action on the _user_ itself rather than on the authorization token. It should at least be better named.

src/pages/PrivateRoute/index.js
```diff
@@ -10,7 +10,7 @@ class PrivateRoute extends Component {
   render() {
     const { children, auth, dispatch } = this.props;
 
-    if (auth.auth.username === 'admin') {
+    if (auth.auth.username) {
```
dbutenhof: Is `auth.auth.username` really right here? Where did we get two `auths` deep in state? I see other components with the same `@connect` pattern (`auth: auth.auth`) that use `auth.username`. What's different here? Or is one or the others wrong?

dbutenhof (commented):
Looks good; a few comments about consistency/style, I think.

src/models/auth.js
aquibbaig: @dbutenhof so, ideally, the logout functionality should hit a server endpoint to gracefully remove the user (also removing the token somehow).

src/pages/PrivateRoute/index.js
aquibbaig: Right, nice catch! Just need to destructure the prop before referencing.

aquibbaig: @dbutenhof should be gone now!

ansible/inventory
```diff
@@ -1,3 +1,2 @@
 [servers]
-staging-server.example.com
-production-server.example.com
+perf30.perf.lab.eng.bos.redhat.com
```
dbutenhof: I don't think we want this change in the external GitHub repo!

public/endpoints.js
```diff
@@ -1,11 +1,9 @@
...
-  result_index: 'vn.result-data-sample.',
-  result_data_index: 'vn.result-data.',
-  pbench_server: 'http://test_domain.com/api/v1',
+  elasticsearch: 'http://elasticsearch.intlab.perf-infra.lab.eng.rdu2.redhat.com:80',
+  results: 'http://pbench.perf.lab.eng.bos.redhat.com:8901',
+  graphql: 'http://graphql.perf.lab.eng.bos.redhat.com:9466/',
+  run_index: 'v6.run.',
+  result_index: 'v5.result-data.',
+  prefix: 'staging-pbench.',
+  pbench_server: 'http://pbench.perf.lab.eng.bos.redhat.com:8901/api/v1',
```
dbutenhof: This also shouldn't go in the repo.

src/models/auth.js
dbutenhof: Well, yes; logout should call the server to remove the authorization session, too. But I was talking more about the fact that logout uses a helper to remove the `localStorage` item while login doesn't. (And the fact that `removeUser` is a misleading name since it's only removing the authorization token.)

src/pages/LoginHandler/index.js
```diff
@@ -42,114 +40,104 @@ class LoginHandler extends React.Component {
...
+    });
+    setAlertOpen(true);
+    if (status === 'success') {
+      // wait for a couple
+      // of seconds before taking the
+      // user to login page.
+      setTimeout(() => {
+        // set logged in to true
+        setLoggedIn({ username });
+      }, 2000);
```
dbutenhof: If login is successful, why are we waiting? And are we really taking the user to the login page on successful login?

public/endpoints.js
aquibbaig: Right, these were changes as a part of the JAM session, will remove

src/pages/LoginHandler/index.js
aquibbaig: So, what I thought initially is to show an alert for a couple of seconds that the login attempt was successful, then redirect to the login page, to provide better user feedback, else as soon as the alert appears, the user is redirected. (probably too fast for users to catch and comprehend)

Also, the function `setLoggedIn` pushes the `/private` page, which will be the Overview page after anisha's PR. (removed the typo in comments now)

src/models/auth.js
aquibbaig: hey @dbutenhof, let me know if the workflow makes a little more sense now

src/services/auth.js
```diff
@@ -0,0 +1,41 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+  });
+};
+
+export const queryLogoutUser = () => {
+  // const endpoint = `${endpoints.pbench_server}/logout`;
+  return { status: 'success', message: 'successfully logged out' };
```
aquibbaig, line 40: should ideally query an API endpoint

src/services/auth.js
```diff
@@ -0,0 +1,41 @@
+import request from '../utils/request';
+
+const { endpoints } = window;
+
+export const queryRegisterUser = async params => {
+  const endpoint = `${endpoints.pbench_server}/user`;
```
dbutenhof, line 6: Isn't this `${endpoints.pbench_server}/register` in the code we merged? Are you working with an old server install?

src/services/auth.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+  });
+};
+
+export const queryLoginUser = params => {
+  const endpoint = `http://127.0.0.1:5000/session`;
```
dbutenhof, line 24: Needs to be `${endpoints.pbench_server}/login`

src/services/auth.js
dbutenhof: I think you should do the server logout before you finish this PR.

src/utils/request.js
```diff
@@ -20,35 +20,78 @@ const codeMessage = {
   504: 'The gateway timed out. ',
 };
 
+const customMessages = new Set([
+  'A user with that name already exists.',
+  'No such user, please register first',
+  'Bad login',
```
dbutenhof, line 26: It bothers me to have the dashboard keying on exact message text; it'll break if we change wording for any reason. (And I presume that at some point we'll be adding message localization.)

A lot of error message responses have an "ID" string that can be more stable... we ought to give that some thought. E.g., the response might be something like this

```
{
    "id": "DUPLICATE_USER",
    "message": "A user with that name already exists."
}
```

and the dashboard could recognize the "known" errors by id and use whatever message text came with it.

Or if you think there's some other server change that would make this easier for you, let's talk about it! Don't treat the server as an immutable external entity; I really want our API to be generally usable for scripting (or even alternate UI integration), but this sort of thing is relevant for any API client anyway.

src/models/auth.js
```diff
@@ -12,9 +14,25 @@ export default {
...
+      return response;
+    },
+    *registerUser({ payload }, { call }) {
+      const response = yield call(queryRegisterUser, payload);
+      return response;
+    },
+    *logoutUser({ payload }, { call }) {
+      const response = yield call(queryLogoutUser, payload);
+      // TODO: Must be removed from the backend through
+      // the response header.
```
dbutenhof, line 29: I'm not sure what "response header" has to do with this; but, yes, we need to make the `logout` call.

What we do on failure is an interesting question; while it would probably be good for the user to know that the authorization token wasn't invalidated, with some sort of error (warning?) popup, we might want to go ahead and remove the browser token to terminate the session?

src/pages/LoginHandler/index.js
```diff
@@ -7,140 +7,136 @@ import {
...
     });
-    if (username !== '' && password !== '') {
-      this.enableSubmitBtn();
-    } else {
-      this.disableSubmitBtn();
+    setAlertOpen(true);
+    if (status === 'success') {
+      // wait for a couple
+      // of seconds before taking the
+      // user to the overview page.
```
dbutenhof, line 61: Thanks; that makes more sense. I suggest expanding this a bit more to clarify that we're waiting to allow the user to notice the successful login alert banner before we continue.

src/pages/SignupHandler/index.js
```diff
@@ -119,13 +133,45 @@ const SignupHandler = props => {
...
+    const { status, message } = response;
+    setAlertComponent({
+      message,
+      variant: status === 'failure' ? 'danger' : 'success',
+    });
+    setAlertOpen(true);
+    if (status === 'success') {
+      // wait for a couple
+      // of seconds before taking the
+      // user to login page.
```
dbutenhof, line 157: Funny, I somehow hadn't noticed the duplication of the comment text; so the login comment was copied from here, where you really _are_ redirecting to the login page, after registration. Ah.

Would it make sense to consolidate the alert and wait into a common method somewhere, though, so the behavior remains consistent... even though the timeout action differs? (It'd be easy enough to pass a one-line lambda in each case, for example.)

src/pages/LoginHandler/index.js
webbnh: Speaking totally for myself, I'm not a fan of ephemeral effects on web pages.  If my attention happens to be pulled away while I'm waiting for the server to respond, I may miss the alert.  In this particular case, that's probably OK, since I'll presumably end up where I expected and wanted to be, anyway...but, that mostly serves to underscore the notion that displaying the alert is not important (and, therefore, perhaps not worth the complication).

I would recommend doing the redirect immediately and then putting a banner (or whatever it's called) on the target page which contains the alert -- that way, it's waiting there for the user when his/her attention returns, and s/he can dismiss the banner once s/he has read it.

src/models/auth.js
webbnh: Logout failure is definitely weird.  If, afterwards, the client and server still have valid, functioning tokens, I would not be in a hurry to dump the client's copy.  From a user perspective, (I think) I have pretty good control over my end of the dialog (e.g., I can close the browser, or reboot the system, if needed...); what I'm concerned about is ensuring that my context on the server does not become available to anyone else.  So, if I cannot log out of the server, I might want to retain my token so that I can try to take some sort of alternative action.

Perhaps we can move forward by dividing the problem.  We need a clear response from the server which indicates whether the token is still valid.  If it's not valid, then, in effect (from the client's perspective) the logout has succeeded, even though the result of the operation was reported as a failure, and we can remove the local copy of the token; however, if the token is still valid, then...I'm not sure what we do, but I don't see how dumping it from the client helps.

src/utils/request.js
aquibbaig: As it is with `umi-request` we need to conditionally compare against a known set of messages for user authentication to handle errors in some other way. So, when we get a response from the server, we need to compare that with some known set of values to render something different that default (throwing exception).

Yes, using an ID with the response can help immensely in this case.

src/services/auth.js
aquibbaig: Currently working on updating it

aquibbaig:
Tracked by #130

PR #127, opened 2021-02-07
Adds Analysis page

anishaswain:
Analysis-Pbench-Dashboard

### DIFF
--- a/config/router.config.js
+++ b/config/router.config.js
@@ -75,6 +75,11 @@ module.exports = [
             component: './Overview',
           },
           {
+            path: '/private/analysis',
+            name: 'analysis',
+            component: './Analysis',
+          },
+          {
             path: '/private/profile',
             name: 'profile',
             exact: true,
--- a/mock/analysis.js
+++ b/mock/analysis.js
@@ -1,0 +1,63 @@
+/* eslint-disable import/prefer-default-export */
+export const donutChart1 = 50;
+
+export const donutChart2 = 70;
+
+export const barGraph1 = [
+  [
+    { name: 'meta1', x: '5-10-2020', y: 100 },
+    { name: 'meta1', x: '6-10-2020', y: 100 },
+    { name: 'meta1', x: '7-10-2020', y: 70 },
+    { name: 'meta1', x: '8-10-2020', y: 10 },
+    { name: 'meta2', x: '9-10-2020', y: 50 },
+    { name: 'meta2', x: '10-10-2020', y: 50 },
+    { name: 'meta2', x: '11-10-2020', y: 30 },
+    { name: 'meta3', x: '12-10-2020', y: 50 },
+    { name: 'meta2', x: '13-10-2020', y: 50 },
+    { name: 'meta2', x: '14-10-2020', y: 30 },
+    { name: 'meta3', x: '15-10-2020', y: 150 },
+    { name: 'meta2', x: '16-10-2020', y: 50 },
+  ],
+  [
+    { name: 'meta1', x: '5-10-2020', y: 100 },
+    { name: 'meta1', x: '7-10-2020', y: 57 },
+    { name: 'meta1', x: '8-10-2020', y: 40 },
+    { name: 'meta1', x: '9-10-2020', y: 20 },
+    { name: 'meta2', x: '10-10-2020', y: 120 },
+    { name: 'meta2', x: '11-10-2020', y: 90 },
+    { name: 'meta2', x: '12-10-2020', y: 0 },
+    { name: 'meta3', x: '13-10-2020', y: 70 },
+    { name: 'meta2', x: '14-10-2020', y: 50 },
+    { name: 'meta2', x: '15-10-2020', y: 30 },
+    { name: 'meta3', x: '16-10-2020', y: 150 },
+  ],
+];
+
+export const barGraph2 = [
+  [
+    { name: 'meta1', x: '5-10-2020', y: 50 },
+    { name: 'meta1', x: '6-10-2020', y: 200 },
+    { name: 'meta1', x: '7-10-2020', y: 70 },
+    { name: 'meta1', x: '8-10-2020', y: 100 },
+    { name: 'meta2', x: '9-10-2020', y: 50 },
+    { name: 'meta2', x: '10-10-2020', y: 50 },
+    { name: 'meta2', x: '11-10-2020', y: 30 },
+    { name: 'meta3', x: '12-10-2020', y: 150 },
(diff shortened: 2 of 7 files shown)


config/router.config.js
```diff
@@ -74,6 +74,11 @@ module.exports = [
             icon: 'dashboard',
             component: './Overview',
           },
+          {
+            path: '/private/analysis',
```
portante, line 78: I am not sure about these `/private` routes.  Don't think we want to provide routes based on their access, but on their function, and then determine the behavior based on access.

src/components/AnalysisChart/index.js
```diff
@@ -0,0 +1,74 @@
...
+          data={{ x: 'Storage capacity', y: usedSpace }}
+          labels={({ datum }) => (datum.x ? `${datum.x}: ${datum.y}%` : null)}
+          thresholds={[{ value: 60 }, { value: 90 }]}
+        />
+      </ChartDonutThreshold>
+    );
+    const bar = (
+      <Chart
+        ariaDesc="mock metdata"
+        ariaTitle="mock meta data"
```
portante, line 42: Is this a mock component or the real version?

src/pages/Analysis/index.js
```diff
@@ -0,0 +1,181 @@
...
+  Card,
+  TextContent,
+  Text,
+  TextVariants,
+  Button,
+} from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import styles from './index.less';
+import AnalysisChart from '@/components/AnalysisChart';
+import { donutChart1, donutChart2, barGraph1, barGraph2 } from '../../../mock/analysis';
```
portante, line 14: Similar question as before, is this a mock component or the real version?

src/common/menu.js
```diff
@@ -27,6 +27,10 @@ const menuData = [
       },
     ],
   },
+  {
+    name: 'Analysis',
+    path: '/private/analysis',
```
portante, line 32: Just another note to consider well the paths we are using.

portante (changes requested):

src/components/AnalysisChart/index.js
anishaswain: This is a chart component is generated with mock data

src/pages/Analysis/index.js
anishaswain: The Analysis page is a real component which contains chart generated with mock data.

src/components/AnalysisChart/index.js
```diff
@@ -0,0 +1,74 @@
...
+        containerComponent={
+          <ChartVoronoiContainer
+            labels={({ datum }) => `${datum.name}: ${datum.y}`}
+            constrai
+            nToVisibleArea
```
dbutenhof: Is this really broken across two lines?

src/pages/Analysis/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+const legends = {
+  'pbench-user-benchmark': 'blue',
+  trafficgen: 'cyan',
+  uperf: 'orange',
+  linpack: 'gold',
+  fio: 'purple',
+};
```
dbutenhof, line 22: Could these be made more dynamic, so the Analysis component doesn't have to be changed for color preferences, or to accommodate additional benchmarks? Does the chart tool have a way to help manage that, for example? I could see color choices being a new item of user-specific metadata in the User db.

src/pages/Analysis/index.js
```diff
@@ -0,0 +1,181 @@
...
+                    <div style={{ height: '260px', width: '260px' }} className={styles.paddingBig}>
+                      <AnalysisChart chartType="donut" usedSpace={donutChart2} />
+                    </div>
+                  </GridItem>
+                </Grid>
+              </GridItem>
+              <GridItem span={5} style={{ borderLeft: '1px solid #b3b2b2' }}>
+                <div className={styles.paddingBig}>
+                  <TextContent>
+                    <Text component={TextVariants.h3}>example.eng.bos.redhat.com</Text>
```
dbutenhof, line 89: Can the hostname here be part of your mock payload rather than hardcoded here? (And below...) Seems like this isn't really at "real component with mock data" since part of the component itself seems to be faked.

src/pages/Analysis/index.js
```diff
@@ -0,0 +1,181 @@
...
+              </TextContent>
+            </div>
+          </Card>
+          <Card className={styles.borderedCard}>
+            <Grid>
+              <GridItem span={8}>
+                <TextContent className={styles.paddingMedium}>
+                  <Text component={TextVariants.h4} style={{ color: 'grey' }}>
+                    {' '}
+                    @metadata.satellite.EC2: filters
```
dbutenhof, line 145: Again, this looks "hardcoded" ... `@metadata.satellite` is a field in the document, but `EC2` is a particular satellite designation, the Elasticsearch "keyword" value of `@metadata.satellite`. It might be useful to filter on data from a particular satellite, for example, but the "EC2" name is specific to our configuration.

src/components/AnalysisChart/index.js
anishaswain: Nope, it should not be. By mistake. Fixed now.

src/pages/Analysis/index.js
portante: Right, but why does the real code reference the `mock` module?

src/components/AnalysisChart/index.js
portante: Right, but why does the real code reference the `mock` module?

portante (changes requested):
Not sure why we have to reference the mock'd data directly from the code.

mock/analysis.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable import/prefer-default-export */
+export const donutChart1 = 50;
+
+export const donutChart2 = 70;
+
+export const barGraph1 = [
+  [
+    { name: 'meta1', x: '5-10-2020', y: 100 },
```
aquibbaig, line 8: Hey Anisha, can we decrease the length of this file by importing these mock objects from a separate file, it would make it much easier to view and review

aquibbaig (changes requested):
Looks good

mock/analysis.js
aquibbaig: I didn't know that I was commenting on the mock file itself, thanks Github :)

aquibbaig (approved):

PR #130, opened 2021-03-15
User authentication PR for the dashboard

aquibbaig:
Follow up PR on #126. #126 was the right way to user authentication but in course of implementing it, we faced a lot of issues with `umi-request`, which made the code hard to debug and review. I do not want us to lose the valuable discussions on that PR, so we will track the user authentication work in this one instead.

TODOS:
- [x] What should be the username? (firstname.toLowercase or the email)
- [x] Remove localhost references (if any)
- [x] This PR introduces breaking changes. All our requests now need to be manually handled, so will discuss if we will include them as a part of this PR (EDIT: This would be done as a part of a new PR)
- [x] General layout component wrapping errors for the authentication layout.

### DIFF
--- a/src/components/GlobalHeader/index.js
+++ b/src/components/GlobalHeader/index.js
@@ -90,7 +90,9 @@ class GlobalHeader extends Component {
                   onSelect={this.onDropdownSelect}
                   isOpen={isProfileDropdownOpen}
                   toggle={
-                    <DropdownToggle onToggle={this.onProfileDropdownToggle}>Admin</DropdownToggle>
+                    <DropdownToggle onToggle={this.onProfileDropdownToggle}>
+                      {username}
+                    </DropdownToggle>
                   }
                   dropdownItems={profileDropdownItems}
                 />
--- a/src/components/GlobalHeader/index.js
+++ b/src/components/GlobalHeader/index.js
@@ -36,12 +36,24 @@ class GlobalHeader extends Component {
     });
   };
 
-  logoutUser = () => {
-    const { dispatch } = this.props;
-    dispatch({
+  logoutUser = async () => {
+    const { dispatch, username } = this.props;
+    const response = await dispatch({
       type: 'auth/logoutUser',
+      payload: username,
     });
-    dispatch(routerRedux.push('/'));
+    const { status, message } = response;
+    if (status === 'failure') {
+      console.error(message);
+    }
+    // We do not want users
+    // to be blocking on any issue
+    // regarding logout.
+    localStorage.removeItem('token');
+    dispatch({
+      type: 'auth/removeUserFromStore',
+      payload: '',
+    }).then(() => dispatch(routerRedux.push(`/`)));
   };
 
   navigateToProfile = () => {
--- a/src/components/GlobalHeader/index.js
+++ b/src/components/GlobalHeader/index.js
@@ -19,7 +19,7 @@ import imgAvatar from '../../assets/avatar.svg';
 import SessionModal from '../SessionModal';
 
 @connect(({ auth }) => ({
-  auth: auth.auth,
+  username: auth.username,
 }))
 class GlobalHeader extends Component {
   constructor(props) {
@@ -55,7 +55,7 @@ class GlobalHeader extends Component {
   };
 
   render() {
-    const { savingSession, dispatch, auth } = this.props;
+    const { savingSession, dispatch, username } = this.props;
     const { isProfileDropdownOpen } = this.state;
 
     const profileDropdownItems = [
@@ -81,7 +81,7 @@ class GlobalHeader extends Component {
               <SearchIcon />
             </Button>
           </PageHeaderToolsItem>
-          {auth.username === 'admin' ? (
+          {username ? (
             <PageHeaderToolsGroup>
               <PageHeaderToolsItem>
                 <Dropdown
--- a/src/components/GlobalHeader/index.js
+++ b/src/components/GlobalHeader/index.js
@@ -36,24 +36,12 @@ class GlobalHeader extends Component {
     });
   };
 
-  logoutUser = async () => {
+  logoutUser = () => {
     const { dispatch, username } = this.props;
-    const response = await dispatch({
+    dispatch({
       type: 'auth/logoutUser',
       payload: username,
     });
-    const { status, message } = response;
-    if (status === 'failure') {
-      console.error(message);
-    }
-    // We do not want users
-    // to be blocking on any issue
-    // regarding logout.
-    localStorage.removeItem('token');
-    dispatch({
-      type: 'auth/removeUserFromStore',
-      payload: '',
-    }).then(() => dispatch(routerRedux.push(`/`)));
   };
 
   navigateToProfile = () => {
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,4 +1,4 @@
-import { queryRegisterUser } from '../services/auth.js';
+import { queryRegisterUser, queryLoginUser } from '../services/auth.js';
 
 export default {
   namespace: 'auth',
@@ -14,8 +14,31 @@ export default {
         payload,
       });
     },
-    // TODO: Add a codeMap for general
-    // purpose error messages.
+    /* eslint-disable camelcase */
+    // TODO: Design a basic codeMap
+    // for general purpose messages.
+    *loginUser({ payload }, { call }) {
+      try {
+        const response = yield call(queryLoginUser, payload);
+        const { message, auth_token } = response;
+        return {
+          message,
+          auth_token,
+          status: 'success',
+        };
+      } catch (error) {
+        const { data } = error;
+        let errortext = 'Something went wrong. Please try again.';
+        if (data) {
+          const { message } = data;
+          errortext = message;
+        }
+        return {
+          message: errortext,
+          status: 'failure',
+        };
+      }
+    },
     *registerUser({ payload }, { call }) {
       try {
         const response = yield call(queryRegisterUser, payload);
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,4 +1,4 @@
-import { queryRegisterUser, queryLoginUser } from '../services/auth.js';
+import { queryRegisterUser, queryLoginUser, queryLogoutUser } from '../services/auth.js';
 
 export default {
   namespace: 'auth',
@@ -42,6 +42,27 @@ export default {
     *registerUser({ payload }, { call }) {
       try {
         const response = yield call(queryRegisterUser, payload);
+        const { message } = response;
+        return {
+          message,
+          status: 'success',
+        };
+      } catch (error) {
+        const { data } = error;
+        let errortext = 'Something went wrong. Please try again.';
+        if (data) {
+          const { message } = data;
+          errortext = message;
+        }
+        return {
+          message: errortext,
+          status: 'failure',
+        };
+      }
+    },
+    *logoutUser({ payload }, { call }) {
+      try {
+        const response = yield call(queryLogoutUser, payload);
         const { message } = response;
         return {
           message,
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -2,7 +2,7 @@ export default {
   namespace: 'auth',
 
   state: {
-    auth: { username: '' },
+    username: '',
   },
 
   effects: {
@@ -24,12 +24,13 @@ export default {
     modifyUser(state, { payload }) {
       return {
         ...state,
-        auth: payload,
+        username: payload,
       };
     },
-    removeUser() {
+    removeUser(state) {
       return {
-        auth: { username: '' },
+        ...state,
+        username: '',
       };
     },
   },
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,3 +1,5 @@
+import { queryRegisterUser } from '../services/auth.js';
+
 export default {
   namespace: 'auth',
 
@@ -12,9 +14,32 @@ export default {
         payload,
       });
     },
-    *logoutUser({ payload }, { put }) {
+    // TODO: Add a codeMap for general
+    // purpose error messages.
+    *registerUser({ payload }, { call }) {
+      try {
+        const response = yield call(queryRegisterUser, payload);
+        const { message } = response;
+        return {
+          message,
+          status: 'success',
+        };
+      } catch (error) {
+        const { data } = error;
+        let errortext = 'Something went wrong. Please try again.';
+        if (data) {
+          const { message } = data;
+          errortext = message;
+        }
+        return {
+          message: errortext,
+          status: 'failure',
+        };
+      }
+    },
+    *removeUserFromStore({ payload }, { put }) {
       yield put({
-        type: 'removeUser',
+        type: 'modifyUser',
         payload,
       });
     },
--- a/src/models/auth.js
+++ b/src/models/auth.js
@@ -1,3 +1,4 @@
+import { routerRedux } from 'dva/router';
 import { queryRegisterUser, queryLoginUser, queryLogoutUser } from '../services/auth.js';
 
 export default {
@@ -5,6 +6,8 @@ export default {
 
   state: {
     username: '',
+    errorMessage: '',
+    successMessage: '',
   },
 
   effects: {
@@ -15,17 +18,20 @@ export default {
       });
     },
     /* eslint-disable camelcase */
-    // TODO: Design a basic codeMap
-    // for general purpose messages.
-    *loginUser({ payload }, { call }) {
+    *loginUser({ payload }, { call, put, all }) {
       try {
         const response = yield call(queryLoginUser, payload);
-        const { message, auth_token } = response;
-        return {
-          message,
-          auth_token,
-          status: 'success',
-        };
+        const { auth_token, username } = response;
+        localStorage.setItem('token', auth_token);
+        yield all([
+          yield put({
+            type: 'loginSuccess',
+            payload: {
+              username,
+            },
+          }),
+          yield put(routerRedux.push('/private')),
+        ]);
       } catch (error) {
         const { data } = error;
         let errortext = 'Something went wrong. Please try again.';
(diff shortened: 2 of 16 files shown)


src/utils/request.js
```diff
@@ -1,54 +1,32 @@
...
-    return;
-  }
-  if (status >= 405 && status < 422) {
-    router.push('/exception/404');
-  }
+  throw error;
 };
 
 const request = extend({
   errorHandler, // extend default error handler with custom actions
```
dbutenhof: Is this really necessary if we're just throwing the error we were passed anyway? What would happen without the errorHandler?

aquibbaig: Hey Dave, so if you navigate to 'src/models/auth' you'll find that we custom handle these errors in our effects. Yes, it is a breaking change which will affect all of our dashboard queries, so you're right in raising that concern

dbutenhof: Ah; right, the errors wouldn't normally be thrown, would they? Is that really the best way to handle this vs looking at HTTP status codes directly... aside from the issues with convincing umi to return the entire Response object, at least. But I suppose that was the motive here?

src/models/auth.js
```diff
@@ -1,3 +1,5 @@
+import { queryRegisterUser, queryLoginUser, queryLogoutUser } from '../services/auth.js';
+
 export default {
   namespace: 'auth',
```
gurbirkalsi, line 5: Is there a specific reason we've initialized the `auth` namespace with: `state: {
    auth: { username: '' },
  }`? Can we consider refactoring `username` as a top level key value pair? In order to avoid confusion with naming, we should avoid including the name of the Redux namespace in the state.

src/components/GlobalHeader/index.js
```diff
@@ -36,12 +36,27 @@ class GlobalHeader extends Component {
...
+    const {
+      dispatch,
+      auth: { username },
+    } = this.props;
+    const response = await dispatch({
       type: 'auth/logoutUser',
+      payload: username,
     });
-    dispatch(routerRedux.push('/'));
+    const { status, message } = response;
```
gurbirkalsi: What if we considered moving the error handling logic to the Redux layer (for example an action named `USER_LOGOUT_FAILURE`) which would update the namespace state with errors received from the server and we can later display in the UI?

src/models/auth.js
```diff
@@ -12,9 +14,76 @@ export default {
...
+      try {
+        const response = yield call(queryLoginUser, payload);
+        const { message, auth_token } = response;
+        return {
+          message,
+          auth_token,
+          status: 'success',
+        };
+      } catch (error) {
+        const { data } = error;
```
gurbirkalsi, line 36: Great initial step for abstracting error handling to the Redux namespace! We'll want to further develop this logic to fire a Redux action that handles populating the namespace state with the error message/status which we can use to visualize in the UI.

gurbirkalsi: @dbutenhof what are your thoughts on having the server as a source of truth for error messages? If so, we can simply pass error messages from the server as an argument instead of creating our own error text here.

src/pages/LoginHandler/index.js
```diff
@@ -7,140 +7,141 @@ import {
...
-    const { username, password } = this.state;
-    if (username === 'admin' && password === 'admin') {
-      this.setLoggedIn({ username });
-    } else {
-      alert('Wrong username/password pair');
-    }
-  };
+  useEffect(
+    () => {
+      if (username.trim() !== '' && password.trim() !== '') {
```
gurbirkalsi: It seems the Submit button is still enabled here while the password and email field is empty after populating it initially. I believe this functionality was working as intended before, is there a specific reason we converted this class component to a functional component?

src/pages/LoginHandler/index.js
```diff
@@ -7,140 +7,141 @@ import {
...
+            style={btnDisabled ? { color: 'black' } : { color: 'white' }}
+          >
+            Submit
+          </Title>
+        </Button>
+      </ActionGroup>
+    </Form>
+  );
+  return (
+    <div>
```
gurbirkalsi: Do we need to return `AuthLayout` in a `<div/>` here?

src/pages/LoginHandler/index.js
```diff
@@ -7,140 +7,141 @@ import {
...
+        />
+      </FormGroup>
+      <ActionGroup>
+        <Button
+          isBlock
+          // variant={variantVal}
+          onClick={handleLoginSubmit}
+          className={styles.btn}
+          id="submitBtn"
+          {...(btnDisabled ? { isDisabled: true } : { isDisabled: false })}
```
gurbirkalsi: What if we refactored this in order to stay consistent with our ternary operator syntax across the codebase?
```
<Button
...other props...
isDisabled={btnDisabled ? true : false}
/>
```

src/utils/request.js
gurbirkalsi: We can forgo the `errorHandler` option completely as `umi-request` is based on the fetch and axios libraries and will allow you to capture errors with simple Promise chaining. We'll handle errors on a case by case on the component level by adopting best practices for error handling with Redux-Saga (https://redux-saga.js.org/docs/basics/ErrorHandling.html).

src/pages/SignupHandler/index.js
```diff
@@ -119,13 +134,47 @@ const SignupHandler = props => {
...
+    setAlertComponent({
+      message,
+      variant: status === 'failure' ? 'danger' : 'success',
+    });
+    setAlertOpen(true);
+    if (status === 'success') {
+      // wait for a couple
+      // of seconds before taking the
+      // user to login page.
+      setTimeout(() => {
```
gurbirkalsi: In our design sync conversations, we mentioned populating the UI masthead with the username and removing the login banner prompt as enough visual feedback to let the user know they've successfully logged in. Can we consider removing this delay as a result?

gurbirkalsi (changes requested):

src/models/auth.js
aquibbaig: I completely agree with that.

src/pages/LoginHandler/index.js
aquibbaig: Hey @gurbirkalsi, I converted this component into a function component to leverage `useEffect` hook, which would allow us to validate the form when either username or password changes. if it were a class component, I'd have to call the validate on a `handleLoginSubmit` and `handlePasswordSubmit`.

src/pages/LoginHandler/index.js
aquibbaig: I agree, removing them

src/pages/SignupHandler/index.js
aquibbaig: @gurbirkalsi That's right, the problem with this is that, if we do not add wait then the toast alert which is a part of this component won't show up (cause the user has already navigated to some other page). I am currently working on having the alert as a part of the parent component `(most likely the AuthLayout/BasicLayout component)`. I will make a custom component if necessary which can be connected to the store for storing error messages as well). So, till that is done, let's keep it as it is. I will add it in the TODOs of this PR.

src/models/auth.js
dbutenhof: @gurbirkalsi I think that we haven't been seriously thinking about the wording and form of error messages yet, but we definitely should since REST clients deserve messages as good as dashboard users. So, yes; when the server returns a failure HTTP status we should always return an error document with a reliable and meaningful `"message"` string that you should be able to present directly. As I mentioned earlier to Aquib, we could also consider producing an "errorcode" token that would uniquely identify a message regardless of arguments or potential future localization in case you need to distinguish errors at a finer granularity than HTTP status.

If we get to the point where we want to produce messages that might benefit from HTML formatting, we can discuss at that time how we want to handle it.

webbnh: I definitely support having the server return an "errorcode" token; however, I would recommend that the conversion of that to a human-readable form be done as close to the human as feasible.  This enables custom formatting, internationalization, and basic de-coupling.  I would counsel against putting this burden on the server -- the presentation of errors belongs in the UI.

Now, you bring up an interesting point with REST clients, but the audience for diagnostics there is different.  For one thing, they can be referred to documentation for translating messages.  Also, they don't have expectations for formatting, and it's reasonable to offer no internationalization in that context.  Mostly, I think we want to avoid creating a temptation for the client to simply pass server message text on to a human user.

aquibbaig: We have used a `codeMap` before. https://github.com/distributed-system-analysis/pbench-dashboard/blob/6e259c44e9f46de7974a73ff3667180594015117/src/utils/request.js#L5. Maybe we can do something similar.

src/pages/LoginHandler/index.js
aquibbaig: > It seems the Submit button is still enabled here while the password and email field is empty after populating it initially. I believe this functionality was working as intended before, is there a specific reason we converted this class component to a functional component?

Thanks! That was a bug, I forgot to look into.

src/models/auth.js
dbutenhof: Client-side message localization requires continued close coordination, especially when there are parameters involved; the server will need to send the raw parameters along with the catalog ID and the "prototype" unlocalized message. There's no reason we can't do that, but it does present additional costs. It does also simplify browser-specific formatting in the messages that we wouldn't want the server to do, so there can be benefits as well. If we're going to go that way, we should target a server error document something like:

```
{
    "message_code": "LOGIN_ERROR",
    "message": "Unable to login user Abominable",
    "parameters": {
        "username": "Abominable"
    }
}
````

(In any case, we should be considering moving to a message catalog within the server, if for no other reason that message grammar and content can be easily reviewed separately, e.g. by documentation, without going through the code to find them all.)

File: src/pages/LoginHandler/index.js [hunk 9]
```diff
@@ -8,139 +8,116 @@ import {
...
-  handlePassWordInputChange = password => {
-    const { username } = this.state;
-    this.setState({
-      password,
+    // shows a alert on failed
+    // login attempt only.
+    // a success response populates data
+    // on the global header and should be
+    // enough of a visual cue.
+    notification.error({
```
aquibbaig: This should be in an `else` block

src/models/auth.js
```diff
@@ -12,24 +17,173 @@ export default {
...
+        yield put({
+          type: 'REGISTER_USER_FAILURE',
+          payload: errortext,
+        });
+      }
+    },
+    *logoutUser({ payload }, { call, put, all }) {
+      try {
+        const response = yield call(queryLogoutUser, payload);
+        const { message } = response;
```
dbutenhof: I don't think we're going to be giving any payload at all on logout, and there's no more "message" field in the success responses. What's this going to do when we stand up a server with Nikhil's updated code?

src/pages/LoginHandler/index.js
```diff
@@ -9,138 +9,90 @@ import {
...
+        />
+      </FormGroup>
+      <ActionGroup>
+        <Button
+          isBlock
+          // variant={variantVal}
+          onClick={handleLoginSubmit}
+          className={styles.btn}
+          id="submitBtn"
+          isDisabled={btnDisabled !== false}
```
dbutenhof: This `!== false` looks weird, and seems to just make the line harder to read. You're initializing it to `true`, and you're only setting it to either `true` or `false`, so is there a reason this can't this just be `btnDisabled`?

At line 87 below, you just use it as a boolean, which makes more sense.

src/models/auth.js
aquibbaig: Right, since we do not use this message in the frontend, we can remove this line.

src/pages/LoginHandler/index.js
aquibbaig: You're right, dunno what I was thinking.

src/models/auth.js
```diff
@@ -87,9 +98,68 @@ export default {
...
+    LOGOUT_USER_SUCCESS(state) {
+      return {
+        ...state,
+        username: '',
+        errorMessage: '',
+        successMessage: '',
+      };
+    },
+    // A failure in logout also removes
+    // the user. We do not want users to be
```
dbutenhof: "removes the user" implies that we're deleting the user rather than merely removing the logged-in username from the Redux state.

dbutenhof (approved):

src/models/auth.js
aquibbaig: fixed it, thanks!

src/pages/SignupHandler/index.js
```diff
@@ -121,7 +120,17 @@ const SignupHandler = props => {
...
   const handleSignupSubmit = () => {
     const { dispatch } = props;
-    dispatch(routerRedux.push('/login'));
+    dispatch({
+      type: 'auth/registerUser',
+      payload: {
+        firstName,
+        lastName,
+        // TODO: Will the username be the email?
+        username: firstName.toLowerCase(),
```
dbutenhof: A while ago I wrote up issue [#2182](https://github.com/distributed-system-analysis/pbench/issues/2182) on the server side, "Reconcile dashboard and server identity management"; we really need to decide exactly what data we're going to manage for a user, what's allowed to change, and how to represent ownership in the Elasticsearch documents using an immutable representation of a user.

One option is to drop "username" and internally use the Postgresql row ID; I don't really like that because it's harder to read. Email isn't good because we definitely want users to be able to change their contact email over time. We might require a text username, make that immutable, and use that internally ... TBD. While "first name" probably isn't a good longer term solution (although, miraculously, our team doesn't have duplicate first names), it's reasonable for now. Another would be, as you say in the comment, to use email. I wouldn't worry about it for this PR, though.

dbutenhof (approved):

src/pages/SignupHandler/index.js
portante: > One option is to drop "username" and internally use the PostgreSQL row ID; I don't really like that because it's harder to read.

Not sure what you mean by harder to read, but in the dashboard we should use email as the handle, and an immutable ID (database row ID).  Email should be verified before used, and once verified, then it has to be unique anyways.  The user can change it, or we can support adding multiple emails per ID.

> Email isn't good because we definitely want users to be able to change their contact email over time.

That is possible using a database ID.  The user can add an email or change it.

> We might require a text username, make that immutable, and use that internally ... TBD.

I am not in favor of this because email has to be unique anyways, and we always have a database ID.

> While "first name" probably isn't a good longer term solution (although, miraculously, our team doesn't have duplicate first names), it's reasonable for now. Another would be, as you say in the comment, to use email. I wouldn't worry about it for this PR, though.

I agree, let's get this PR to land, we can change to email in a future PR.

portante (approved):

src/models/auth.js
```diff
@@ -12,24 +17,172 @@ export default {
...
+    *removeSuccessMessage({ payload }, { put }) {
+      yield put({
+        type: 'modifySuccessMessage',
         payload,
       });
     },
   },
 
   reducers: {
+    LOGIN_USER_SUCCESS(state, { payload }) {
```
gurbirkalsi: Lets consider maintaining camelCase naming conventions as mentioned by Redux best practices: https://redux.js.org/style-guide/style-guide/#priority-c-rules-recommended

gurbirkalsi (changes requested):
Looks great, one last small consideration!

src/pages/LoginHandler/index.js
```diff
@@ -9,138 +9,89 @@ import {
...
+const mapStateToProps = state => {
+  const { auth } = state;
+  return { auth };
+};
 
-  componentDidMount = () => {
-    this.disableSubmitBtn();
-  };
+const LoginHandler = props => {
+  const [username, setUsername] = useState('');
```
gurbirkalsi, line 22: Nice use of hooks!

gurbirkalsi (approved):

gurbirkalsi:
Once #124 is merged, we can rebase and resolve e2e test environment errors before merging. Great work @aquibbaig!

src/pages/LoginHandler/index.js
```diff
@@ -9,138 +9,89 @@ import {
...
-  };
-
-  disableSubmitBtn = () => {
-    this.setState({
-      variantVal: 'tertiary',
-      btnColor: 'black',
-      btnDisabled: 'true',
+      type: 'auth/loginUser',
+      payload: {
+        username,
```
gurbirkalsi, line 43: Are we relying on the server to validate emails as part of the login endpoint? We can perform validation on the client in order to reduce the number of bad requests making its way to the server and improve the load put on our API. We can make use of the `validateEmail` utility at `./utils/validator` here.

src/pages/Results/index.js
```diff
@@ -178,6 +176,25 @@ class Results extends Component {
         accessor: 'end',
         Cell: row => <span>{getDiffDate(row.value)}</span>,
       },
+      {
+        Header: 'Favorited',
+        accessor: 'favorited',
+        show: username === '',
```
gurbirkalsi: Did you mean to say `username !== ''` here in order to check for a logged in user?

gurbirkalsi (changes requested):

src/pages/Results/index.js
aquibbaig: right, sorry! I was testing stuff

src/pages/LoginHandler/index.js
aquibbaig: Just fixed, thanks! @gurbirkalsi

src/pages/LoginHandler/index.js
```diff
@@ -9,138 +9,102 @@ import {
...
+  const [btnDisabled, setBtnDisabled] = useState(true);
 
-  setLoggedIn = payload => {
-    const { dispatch } = this.props;
-    dispatch({
-      type: 'auth/loadUser',
-      payload,
+  const handleUsernameChange = val => {
+    setUsername(val);
+    const validEmail = validateEmail(val);
```
gurbirkalsi, line 31: We should consider disabling the submit button to prevent bad requests going to the API anytime the email is invalid. We perform a similar check for form validation for the sign up flow that we can adopt here.

gurbirkalsi (changes requested):

src/pages/LoginHandler/index.js
aquibbaig: Ah, that's right! Thanks for pointing that out. Just made the changes @gurbirkalsi

PR #131, opened 2021-03-17
Improved mock generation with mocked network request responses

gurbirkalsi:
In order to support continued dashboard component development while the transition to the `pbench_server` is underway, the dashboard requires an advanced mock environment that allows for generation of response data and interception of network requests. By running the the dashboard with `yarn start:mock`, the mock development environment starts and listens for requests to intercept as defined in `mock/api.js`. This PR introduces a "fake data generator" library called casual to mock various response fields.

As server request postprocessing progress continues, the mock responses defined in `mock/api.js` will become significantly smaller. The primary motivation behind mocking requests is to allow for rapid prototyping of React components throughout the dashboard and other various feature development. For more details on the mock development environment, please reference the updated `README.md`.

### DIFF
--- a/mock/api.js
+++ b/mock/api.js
@@ -1,84 +1,253 @@
-import constants from '../config/constants';
+import moment from 'moment';
+import casual from 'casual';
 
-// Mocked test index components
-const endpoints = {
-  prefix: 'test_prefix.',
-  run_index: 'vn.run-data.',
-  run_toc_index: 'vn.run-toc.',
-  result_index: 'vn.result-data-sample.',
-  result_data_index: 'vn.result-data.',
-};
+const DEFAULT_SIZE = 100;
 
-// Generate controllers as per max page size options
-const maxTableSize = parseInt(constants.tableSizeOptions.pop(), 10);
-let generatedBuckets = new Array(maxTableSize).fill({});
-generatedBuckets = generatedBuckets.map((val, index) => {
-  const key = index + 1;
-  return {
-    key: `controller_${key}`,
-    controller: `controller_${key}`,
-    results: key,
-    last_modified_value: key,
-    last_modified_string: key.toString(),
-  };
-});
-export const generateMockControllerAggregation = generatedBuckets;
+export const mockIndices = new Array(DEFAULT_SIZE).fill().map(() => moment().format('YYYY-MM'));
 
-const prefix = endpoints.prefix + endpoints.run_index.slice(0, -1);
-export const mockIndices = ['2019-09', '2019-08'];
+export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index) => ({
+  controller: casual.word,
+  key: casual.word,
+  last_modified_string: moment.utc() + index,
+  last_modified_value: moment.utc() + index,
+  results: casual.integer(1, DEFAULT_SIZE),
+}));
 
 export const mockResults = {
+  took: DEFAULT_SIZE,
+  hits: {
+    total: {
+      value: DEFAULT_SIZE,
+    },
+    hits: new Array(DEFAULT_SIZE).fill().map((value, index) => ({
+      _index: moment().format('YYYY-MM'),
+      _type: casual.word,
+      _id: casual.uuid,
+      _score: null,
+      _source: {
+        '@metadata': {
+          controller_dir: casual.word,
+        },
+        run: {
+          controller: casual.word,
+          name: casual.word,
+          start: moment.utc() + index,
+          end: moment.utc() + index,
+          id: casual.uuid,
+          config: casual.word,
+        },
+      },
+      sort: [casual.unix_time],
+    })),
+  },
+};
+
+export const mockSamples = {
+  _scroll_id: casual.uuid,
+  hits: {
+    total: DEFAULT_SIZE,
+    hits: new Array(DEFAULT_SIZE).fill().map((value, index) => ({
+      _source: {
+        '@timestamp': moment.utc(),
+        run: {
+          id: 'test_id',
+          controller: 'test_controller',
+          name: 'test_result',
+          script: casual.random_element(['fio', 'uperf']),
+          date: moment.utc() + index,
+          start: moment.utc() + index,
+          end: moment.utc() + index,
+          user: casual.username,
+        },
+        iteration: {
+          name: `iteration${index + 1}`,
+          number: index + 1,
+        },
+        benchmark: {
+          max_stddevpct: casual.integer(1, 5),
+          primary_metric: casual.random_element(['clat', 'iops_sec']),
+          uid: 'benchmark_name:fio-controller_host:test_controller',
+          name: casual.random_element(['fio', 'uperf']),
+          uid_tmpl: 'benchmark_name:%benchmark_name%-controller_host:%controller_host%',
+        },
+        sample: {
+          client_hostname: casual.random_element(['1', '2']),
+          closest_sample: casual.integer(1, 5),
+          description: casual.short_description,
+          mean: casual.integer(1, DEFAULT_SIZE * 100),
+          role: 'client',
+          stddev: casual.integer(1, DEFAULT_SIZE * 100),
+          stddevpct: casual.integer(1, DEFAULT_SIZE),
+          uid: casual.random_element(['client_hostname:1', 'client_hostname:2']),
+          measurement_type: casual.random_element(['latency', 'throughput']),
+          measurement_idx: casual.integer(0, 4),
+          measurement_title: casual.random_element(['clat', 'iops_sec']),
+          uid_tmpl: 'client_hostname:%client_hostname%',
+          '@idx': casual.integer(0, 4),
+          name: `sample${casual.integer(1, 5)}`,
+          start: moment.utc() + index,
+          end: moment.utc() + index,
+        },
+      },
+    })),
+  },
+  aggregations: {
+    id: {
+      buckets: [
+        {
+          key: 'test_id',
+          type: {
+            buckets: [
+              {
+                key: 'latency',
+                title: {
+                  buckets: [
+                    {
+                      key: 'clat',
+                      uid: {
+                        buckets: new Array(2).fill().map((value, index) => ({
+                          key: `client_hostname:${index + 1}`,
+                        })),
+                      },
+                    },
+                  ],
+                },
+              },
+              {
+                key: 'throughput',
+                title: {
+                  buckets: [
+                    {
+                      key: 'iops_sec',
+                      uid: {
+                        buckets: new Array(2).fill().map((value, index) => ({
+                          key: `client_hostname:${index + 1}`,
+                        })),
+                      },
+                    },
+                  ],
+                },
+              },
+            ],
+          },
+        },
+      ],
+    },
+    name: {
+      buckets: [
+        {
+          key: 'test_result',
+        },
+      ],
+    },
+    controller: {
+      buckets: [
+        {
+          key: 'test_controller',
+        },
+      ],
+    },
+  },
+};
+
+export const mockDetail = {
   hits: {
     hits: [
       {
         _source: {
           run: {
-            name: 'a_test_run',
-            start: '1111-11-11T11:11:11+00:00',
-            id: '1111',
-            end: '1111-11-11T11:11:12+00:00',
-            controller: 'test_run.test_domain.com',
-            config: 'test_size_1',
+            id: 'test_id',
+            controller: 'test_controller',
+            name: 'test_result',
+            script: casual.random_element(['fio', 'uperf']),
+            date: moment.utc(),
+            start: moment.utc(),
+            end: moment.utc(),
+            user: casual.username,
           },
-          '@metadata': {
-            controller_dir: 'test_run.test_domain.com',
-          },
-        },
-      },
-      {
-        _source: {
-          run: {
-            name: 'b_test_run',
-            start: '1111-11-11T11:11:13+00:00',
-            id: '2222',
-            end: '1111-11-11T11:11:14+00:00',
-            controller: 'b_test_run.test_domain.com',
-            config: 'test_size_2',
-          },
-          '@metadata': {
-            controller_dir: 'b_test_run.test_domain.com',
-          },
+          host_tools_info: [
+            {
+              hostname: 'test_host',
+              tools: {
+                iostat: '--interval=3',
+                mpstat: '--interval=3',
+                perf: "--record-opts='record -a --freq=100'",
+                pidstat: '--interval=30',
+                'proc-interrupts': '--interval=3',
+                'proc-vmstat': '--interval=3',
+                sar: '--interval=3',
+                turbostat: '--interval=3',
+              },
+            },
+          ],
         },
       },
     ],
   },
 };
 
+export const mockTableContents = {
+  hits: {
+    total: DEFAULT_SIZE,
+    hits: new Array(DEFAULT_SIZE).fill().map((value, index) => ({
+      _source: {
+        parent: `${casual.word}/${casual.word}/${casual.word}/${casual.word}/${casual.word}/${
+          casual.word
+        }`,
+        directory: `${casual.word}/${casual.word}/${casual.word}/${casual.word}/${casual.word}/${
+          casual.word
+        }`,
+        mtime: moment.utc() + index,
+        mode: casual.word,
+        name: casual.word,
+        ancestor_path_elements: [
+          casual.word,
+          casual.word,
+          casual.word,
+          casual.word,
+          casual.word,
+          casual.word,
+        ],
+        files: [
+          {
+            name: casual.uid,
+            mtime: moment.utc() + index,
+            size: casual.integer(1, DEFAULT_SIZE),
+            mode: '0o644',
+            type: casual.random_element([
+              'reg',
+              'areg',
+              'lnk',
+              'sym',
+              'dir',
+              'fifo',
+              'cont',
+              'chr',
+              'blk',
+              'spr',
+            ]),
+          },
+        ],
+      },
+    })),
+  },
+};
+
 export const mockMappings = {
-  [`${prefix}.2019-09`]: {
+  test_index: {
     mappings: {
       properties: {
         run: {
           properties: {
-            config: { type: 'keyword' },
-            name: { type: 'keyword' },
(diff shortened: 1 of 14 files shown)


src/utils/parse.test.js
```diff
@@ -1,26 +1,20 @@
...
 
 describe('test generateClusters', () => {
-  const parsedClusterData = generateClusters(
-    expectedSampleData.runs,
-    expectedSampleData.iterationParams
-  );
   it('should generate cluster data', () => {
     expect(Object.keys(parsedClusterData).length > 0);
   });
-  it('should equal mocked cluster data', () => {
```
gurbirkalsi, line 23: Please note we'll want to consider a new way for comparing mocked outcomes from our postprocessing without having the massive parsed response defined in our mock files. Eventually, this definition will move to the server as the server will handle postprocessing before sending a simplified response to the dashboard.

src/services/dashboard.js
```diff
@@ -133,7 +137,9 @@ export async function queryTocResult(params) {
     selectedDateRange
   )}/_search?q=run_data_parent:"${id}"`;
 
-  const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+  const endpoint = MOCK_UI
```
gurbirkalsi, line 140: As a first step, we'll define custom endpoints that have not yet been defined in `pbench_server`. Eventually, these will be replaced with their appropriate endpoints. These mocked endpoints are only available when the `MOCK_UI` env variable is set to True.

anishaswain:
@gurbirkalsi I am experiencing an error such as
 ```
ExpandableTable.js: Uncaught TypeError: data.some is not a function
```
while trying to access dashboard with `yarn start:mock`. Any idea how to resolve the same?

**update**: _got resolved automatically with a few more try. Not sure how._

File: mock/api.js [hunk 3]
```diff
@@ -1,381 +1,299 @@
...
-              },
-              run: {
-                id: 'test_run_id',
-                controller: 'test_controller',
-                name: 'test_run_name',
-                script: 'test_script',
-                config: 'test_config',
-              },
+export const mockMappings = {
+  text_index: {
```
aquibbaig: We look for 
```
const index = endpoints.prefix + endpoints.run_index + indices[0];
```
in `search/fetchIndicexMapping` which according to the mock endpoints file becomes `test_prefix.v.run-data.2021-03` but the key given here is `test_index` which doesn't fetch mappings on to the search page.

aquibbaig (commented):
Thanks for this PR @gurbirkalsi. It will be really helpful in running the dashboard in a mock environment.

mock/api.js
gurbirkalsi: Thanks for catching this Aquib! This is an interesting edge case where we're unable to build the correct index due to api.js being referenced as part of the Node build process which means we don't have access to `window.endpoints`. I'm including a temporary fix that allows us to reference `test_index` when in the mock development environment. We'll want to work on adapting the server endpoint for fetching index mappings to take an index parameter so that the dashboard doesn't parse the response for a specific index name.

gurbirkalsi:
e2e tests are breaking due to the testing environment referencing our new mocked response generators. I will address the e2e tests once PR #124 is merged as most of the selectors for each page will change. Let me know if there are any other pending issues with this PR!

aquibbaig (approved):
LGTM

src/services/sessions.js
```diff
@@ -24,7 +27,9 @@ export async function getSession(params) {
 
 // queries all the available shared sessions from the database to display
 export async function getAllSessions() {
-  const endpoint = `${endpoints.pbench_server}/graphql`;
+  const endpoint = MOCK_UI
+    ? `${endpoints.pbench_server}/sessions/list`
+    : `${endpoints.pbench_server}/graphql`;
```
portante, line 32: Why not just usupr the `/graphql` endpoint for the mock'd environment as well?

portante (dismissed):
Seems like we want to avoid knowledge of the mock'd environment as much as possible.  What can we do to avoid exposing that?

mock/api.js
```diff
@@ -1,381 +1,297 @@
...
-      },
-      name: {
-        buckets: [
+        files: [
           {
-            key: 'test_run_name',
+            name: casual.uid,
+            mtime: moment.utc() + index,
+            size: casual.integer(1, DEFAULT_SIZE),
+            mode: casual.word,
```
dbutenhof: Didn't Anisha have some outstanding code to _parse_ the mode string (once the effect actually returns it)? Is it worth building a better fake (octal numeric) mode now (or just hardcoding something like `"0o644"`?) vs deferring that to later?

mock/api.js
```diff
@@ -1,381 +1,297 @@
...
+            type: casual.word,
```
dbutenhof: Similar comment here: should we worry about dashboard code that cares about the file type "keywords" returned from Elasticsearch and be more careful here about mocking? Maybe just hardcoded `"reg"` would be good enough?

mock/api.js
```diff
@@ -1,381 +1,297 @@
...
+            config: { type: 'string', index: 'not_analyzed' },
+            controller: { type: 'string', index: 'not_analyzed' },
+            date: { type: 'date', format: 'dateOptionalTime' },
+            end: { type: 'date', format: 'dateOptionalTime' },
+            id: { type: 'string', index: 'not_analyzed' },
+            iterations: { type: 'string' },
+            name: { type: 'string', index: 'not_analyzed' },
+            start: { type: 'date', format: 'dateOptionalTime' },
+            user: { type: 'string', index: 'not_analyzed' },
           },
```
dbutenhof, line 251: Not sure how much this'll matter for e2e testing, for for the record most of the properties are `"keyword"` now. For reference, here's an actual Postman response to the `_mappings` query:

```
{
  "drb.v6.run-data.2020-04": {
    "mappings": {
      "_meta": {
        "version": "6"
      },
      "date_detection": false,
      "properties": {
        "@generated-by": {
          "type": "keyword"
        },
        "@metadata": {
          "properties": {
            "controller_dir": {
              "type": "keyword"
            },
            "file-date": {
              "type": "date"
            },
            "file-name": {
              "type": "keyword"
            },
            "file-size": {
              "type": "long"
            },
            "md5": {
              "type": "keyword"
            },
            "pbench-agent-version": {
              "type": "keyword"
            },
            "raw_size": {
              "type": "long"
            },
            "result-prefix": {
              "type": "text"
            },
            "satellite": {
              "type": "keyword"
            },
            "tar-ball-creation-timestamp": {
              "type": "date"
            },
            "toc-prefix": {
              "type": "text"
            }
          }
        },
        "@timestamp": {
          "type": "date"
        },
        "authorization": {
          "properties": {
            "access": {
              "type": "keyword"
            },
            "owner": {
              "type": "keyword",
              "null_value": "NONE"
            }
          }
        },
        "host_tools_info": {
          "type": "nested",
          "properties": {
            "hostname": {
              "type": "keyword"
            },
            "hostname-f": {
              "type": "keyword"
            },
            "hostname-s": {
              "type": "keyword"
            },
            "label": {
              "type": "keyword"
            },
            "tools": {
              "properties": {
                "disk": {
                  "type": "text"
                },
                "haproxy-ocp": {
                  "type": "text"
                },
                "iostat": {
                  "type": "text"
                },
                "mpstat": {
                  "type": "text"
                },
                "oc": {
                  "type": "text"
                },
                "perf": {
                  "type": "text"
                },
                "pidstat": {
                  "type": "text"
                },
                "pprof": {
                  "type": "text"
                },
                "proc-interrupts": {
                  "type": "text"
                },
                "proc-vmstat": {
                  "type": "text"
                },
                "prometheus-metrics": {
                  "type": "text"
                },
                "sar": {
                  "type": "text"
                },
                "turbostat": {
                  "type": "text"
                },
                "vmstat": {
                  "type": "text"
                }
              }
            }
          }
        },
        "run": {
          "properties": {
            "config": {
              "type": "keyword"
            },
            "controller": {
              "type": "keyword"
            },
            "date": {
              "type": "date"
            },
            "end": {
              "type": "date"
            },
            "id": {
              "type": "keyword"
            },
            "iterations": {
              "type": "text"
            },
            "name": {
              "type": "keyword"
            },
            "script": {
              "type": "keyword"
            },
            "start": {
              "type": "date"
            },
            "toolsgroup": {
              "type": "keyword"
            },
            "user": {
              "type": "keyword"
            }
          }
        },
        "sosreports": {
          "type": "nested",
          "properties": {
            "hostname-f": {
              "type": "keyword"
            },
            "hostname-s": {
              "type": "keyword"
            },
            "inet": {
              "type": "nested",
              "properties": {
                "ifname": {
                  "type": "keyword"
                },
                "ipaddr": {
                  "type": "ip"
                }
              }
            },
            "inet6": {
              "type": "nested",
              "properties": {
                "ifname": {
                  "type": "keyword"
                },
                "ipaddr": {
                  "type": "keyword"
                }
              }
            },
            "md5": {
              "type": "keyword"
            },
            "name": {
              "type": "keyword"
            },
            "sosreport-error": {
              "type": "text"
            }
          }
        }
      }
    }
  }
}
```

mock/api.js
```diff
@@ -1,381 +1,297 @@
...
+  'GET /controllers/months': mockIndices,
+  'POST /controllers/list': mockControllers,
+  'POST /datasets/list': mockResults,
+  'POST /datasets/detail': mockDetail,
+  'POST /datasets/toc': mockTableContents,
+  'POST /datasets/samples': mockSamples,
+  'POST /mappings': mockMappings,
+  'POST /search': mockSearch,
+  'POST /sessions/list': mockSessions,
+  'POST /sessions/create': mockSession,
```
dbutenhof, line 307: How are these URIs being interpreted? The server `/datasets/list` and `datasets/detail` APIs aren't in the server main yet; and while `/datasets/toc` and `/datasets/samples` are reasonable suggestions those server APIs aren't implemented; nor are native search and "sessions" (if such things will even exist in the current terms, which seems unlikely).

I guess I need to search to see where this is referenced, though; perhaps it'll make sense.

src/pages/document.ejs
```diff
@@ -9,7 +9,7 @@
   <link rel="icon" href="<%= context.config.publicPath +'favicon.ico'%>" type="image/x-icon">
   
   <!-- runtime endpoints config -->
-  <script src="<%= context.config.publicPath +'endpoints.js'%>"></script>
+  <script src="<%= context.config.publicPath + context.config.define.ENDPOINTS_ENV%>"></script>
```
dbutenhof, line 12: FYI, I'm working on turning this into something like:

```
  <script type="text/javascript">
    fetch(`${window.origin}/api/v1/endpoints`)
      .then(function(response) {
        if (!response.ok) {
          console.log(`ERROR: ${response.statusText}: ${response.statusText}`);
        } else {
          response.json().then(function(data) {
            window.endpoints = data;
          })
        }
      });
  </script>
```

Although I'm not convinced that's the right packaging and I need to dump the `console.log` and invert that `if`. (Though this morning I've managed to break it every time I simplify anything.)

src/services/dashboard.js
```diff
@@ -53,8 +54,9 @@ export async function queryResults(params) {
       endpoints.run_index,
       selectedDateRange
     )}/_search`;
-
-    const endpoint = `${endpoints.pbench_server}/elasticsearch`;
+    const endpoint = MOCK_UI
+      ? `${endpoints.pbench_server}/datasets/list`
+      : `${endpoints.pbench_server}/elasticsearch`;
```
dbutenhof, line 59: Ah; interesting. Better to mock based on distinct endpoints rather than the klunky old JSON patterns from the payload; and eventually we can just simplify this to drop the elasticsearch alternative. I like.

src/services/search.js
```diff
@@ -32,35 +37,35 @@ export async function searchQuery(params) {
...
-                      gte: selectedDateRange.start,
-                      lte: selectedDateRange.end,
-                    },
-                  },
-                },
-                must: {
-                  query_string: {
-                    query: `*${query}*`,
-                    analyze_wildcard: true,
+          size: 10000,
```
dbutenhof, line 40: Wow; GitHub's rendition of this indentation change is really bizarre.

src/services/sessions.js
dbutenhof: Because the old mocks depended on JSON payload; the new ones are strictly URI endpoints, which is a lot cleaner and more reliable!

dbutenhof (approved):
This looks great. Some minor concern about the mocks of the file-level detail in the TOC, but I believe that doesn't impact the dashboard code until one of Anisha's outstanding PRs hits. (And currently that data, I believe, isn't even making it through the postprocessing effect.)

src/services/sessions.js
gurbirkalsi: Correct! We would be able to remove all references to the `MOCK_UI` once everything routes through the `pbench_server` endpoint. This is temporary solution in order to allow us to use the mock environment as soon as possible. Ideally, we can specify a single endpoint pointing to the `pbench_server` regardless of being in a mocked environment or not.

mock/api.js
gurbirkalsi: I'm certainly open to including a hardcoded value or creating a small helper to generate fake modes. @anishaswain Can you confirm if you're working on parsing the mode string as part of your work on table of contents data?

mock/api.js
gurbirkalsi: Thanks for pointing this out! Is there a definition for type of "keywords" we can expect from Elasticsearch? In that case, we can include a solution that randomly selects a keyword type from the list for each mocked object.

mock/api.js
gurbirkalsi: Interesting! I was referencing responses from the dashboard deployed in production which might be referencing an older version of Elasticsearch. For now, this doesn't cause problems because the dashboard is only concerned with seeing a string as part of the response raw value. It seems "keyword" and "text" equate to a string as part of the final raw value. Please correct me if I'm wrong here.

mock/api.js
gurbirkalsi: You'll notice in our `services/` definitions we specify `MOCK_UI` which is a runtime environment variable that allows us to determine if the dashboard is being run in a mocked environment. If that is the case, we tell the dashboard to hit these "fake" endpoints which ultimately return the mocked response. Over time, we can substitute these "fake" endpoints with real endpoints to `pbench_server` so we don't have to say `MOCK_UI ? /sessions/list : /graphl` when we build our endpoint url before submitting a request.

src/pages/document.ejs
gurbirkalsi: Sweet! Looking forward to being able to query the endpoint configuration on demand.

mock/api.js
dbutenhof: It's a lookup from the tar file type codes, in `indexer.py`:

```
    _mode_table = _dict_const(
        [
            (tarfile.REGTYPE, "reg"),
            (tarfile.AREGTYPE, "areg"),
            (tarfile.LNKTYPE, "lnk"),
            (tarfile.SYMTYPE, "sym"),
            (tarfile.DIRTYPE, "dir"),
            (tarfile.FIFOTYPE, "fifo"),
            (tarfile.CONTTYPE, "cont"),
            (tarfile.CHRTYPE, "chr"),
            (tarfile.BLKTYPE, "blk"),
            (tarfile.GNUTYPE_SPARSE, "spr"),
        ]
    )
```

mock/api.js
dbutenhof: Yes, in ES1, we used `"string"` for most values. In ES7, indexed values are `"keyword"` and free-search text values (we don't have that many) are `"text"`.

Indeed, the dashboard mocking only needs to worry about what the dashboard cares about, and I don't believe it really cares about the type, only field names. Just thought I'd point this out... and aware that I posted way too much of the sample for that purpose. ;-)

src/pages/document.ejs
dbutenhof: Actually, I was wondering whether you'd object philosophically to including that sort of code in the "root loader" html file. This is encouraging.

However, since you didn't object outright... this works beautifully on my VM, but I'm having problems in the e2e environment. The `fetch` is returning the full index.html, and the `.json()` is throwing because of the initial `<` character. (That's why I had the `console.log()` statements in here; although to actually nail this down I had to go to `response.text()` and print that before trying to manually jsonify it.) This may be partly because this is running too early, before the request mocking is set up, but getting the page html from the yarn server instead of a 404 still confuses me. Why would it ignore the path? Do you have any ideas here? Is there some yarn configuration, for example, that's making the development server return the page html for _every_ `GET` until we set up the mock overrides???

anishaswain (approved):
LGTM once the tests pass. Good work @gurbirkalsi

src/pages/document.ejs
gurbirkalsi: My only concern is the added latency and overhead introduced in querying for a file before loading the dashboard bundle. Since the endpoint config is very small in size we can assume minimal latency and implement strategies for caching the definition in the future. 

Have you tried setting `MOCK=true` in the e2e environment? We currently do this in `package.json` for starting the dashboard in mocked environment: `"start:mock": "cross-env UMI_UI=none MOCK=true umi dev"`. It also may be helpful to debug the environment by setting the e2e tests to run without the headless mode: 

```
browser = await puppeteer.launch({
    headless: true -> false, // headless should be set to false 
    args: ['--no-sandbox'],
  });
```

dbutenhof: In terms of HTTP protocol and Javascript, this is essentially identical to what you had before; a GET back to the origin, resulting in a JSON document, which is stashed in the `window` object for universal access. We do have the reverse proxy to the Pbench server instead of Apache's built-in file server, but we also have no disk access. I figure the performance and latency differences are going to be too small to measure.

Beyond that... it really seems like my problems are more a matter of how the npm dev server runs & is configured rather than in the Javascript itself. I tried copying your `start:mock` into my environment, but that doesn't appear to make any difference.

dbutenhof (approved):

mock/api.js
gurbirkalsi: I've updated the mock generator for the `type` field here to randomly select a tar file type upon generation.

mock/api.js
gurbirkalsi: I've updated the `mode` field here to be a hardcoded value for "0o644".

gurbirkalsi:
Merging the PR in its current state, e2e tests will be addressed with selector changes in PR #124

PR #135, opened 2021-03-30
Remove unused jspdf dependency from node modules

gurbirkalsi:
This PR addresses the following Dependabot alert: https://github.com/distributed-system-analysis/pbench-dashboard/security/dependabot/package.json/jspdf/open

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "enzyme-adapter-react-16": "^1.15.1",
     "html2canvas": "^1.0.0-alpha.12",
     "jschart": "^1.0.7",
-    "jspdf": "^1.5.3",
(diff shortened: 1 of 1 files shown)


portante (changes requested):
Looks good, just needs some passing tests ...

aquibbaig:
@portante I'm afraid the tests are failing due to mock generation PR. Asserting randomly generated values is a bit tricky. We should probably use a different approach with our e2e tests. (like "table data should not be null", "date picker component months should not be null", etc). I think we have to do that in a separate PR.

aquibbaig (approved):
Looks good!

portante (approved):

Issue #136, opened 2021-03-31
Fix dashboard e2e tests


aquibbaig:
e2e tests are failing after PR #131. That's probably because the mock data generated by `casual.js` is hard to assert during the test environment. We should look at a different approach to our e2e tests. We could check for not null values in our page components as a starting point

portante:
Fixed by #138.

PR #139, opened 2021-03-31
fix e2e tests for controllers page

aquibbaig:
Initial PR aimed at fixing #136.

Some important things to note:
- I found that `casual.js` does not generate unique data for a specific entity. For example, if we generate 100 controller names, I have noticed about 5-10 collisions in the name. The problem is that sometimes they have the same key, and if they do antd's table sorting feature would not work. Hence I have commented out that part of the code.
- I have taken a newer approach to test as to select a record beforehand which is expected, then do some action and assert if the result and earlier record are the same. Since it is impossible to assume generated data beforehand and assert it.

### DIFF
--- a/src/e2e/controllers.e2e.js
+++ b/src/e2e/controllers.e2e.js
@@ -47,16 +47,20 @@ describe('controller page component', () => {
       const testController = await page.$eval('.ant-table-row', elem =>
         elem.getAttribute('data-row-key')
       );
-      expect(testController).toBe('controller_1');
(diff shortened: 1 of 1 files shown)


gurbirkalsi:
Hey @aquibbaig just a heads up that these selectors will most likely change very soon with PR #124 as the table component will now use PatternFly classNames/ids. I've started working on updating the selectors with the mock data we generate as part of the new environment.

aquibbaig:
Fixed under #124

PR #141, opened 2021-04-07
Endpoints

### LINKED ISSUE #107: Use server `/api/v1/endpoints` API to to auto-configure the Dashboard
Improve dashboard deployment experience by using the Pbench server `/api/v1/endpoints` API instead of a hardcoded `endpoints.js` file built into the javascript bundle loaded on the server.

That way we can build one dashboard bundle and test it on staging and move it to production without having to change or rebuild the bundle.

_Originally posted by @portante in https://github.com/distributed-system-analysis/pbench-dashboard/issues/106#issuecomment-737425234_

dbutenhof:
Update the dashboard code to use the Pbench server `/api/v1/endpoints` API to self-configure without requiring a static `endpoints.js` file for prefix, index names, and end point URLs. This makes a deployed dashboard fully self-configuring to match the co-located Pbench server.

For example, a containerized Pbench server could include the dashboard code and you could simply connect to `/dashboard` on the external container port, and everything would work.


config/config.js
```diff
@@ -4,7 +4,7 @@ export default {
   // Note: typeof(process.env.MOCK) === string
   define: {
     MOCK_UI: process.env.MOCK === 'true',
-    ENDPOINTS_ENV: process.env.MOCK === 'true' ? 'endpoints.mock.js' : 'endpoints.js',
+    ENDPOINTS_ENV: process.env.MOCK === 'true' ? '/mock/endpoints.mock.json' : '/api/v1/endpoints',
```
dbutenhof: Oops; `ENDPOINTS_ENV` is no longer used or useful; I should remove it.

src/pages/document.ejs
```diff
@@ -9,7 +9,21 @@
   <link rel="icon" href="<%= context.config.publicPath +'favicon.ico'%>" type="image/x-icon">
   
   <!-- runtime endpoints config -->
-  <script src="<%= context.config.publicPath + context.config.define.ENDPOINTS_ENV%>"></script>
+  <script type="text/javascript">
+    fetch( '/api/v1/endpoints' )
+      .then(response => {
+        console.log(`RESPONSE: ${response.status},${response.statusText}`);
```
dbutenhof: Obviously the `console.log` statements here will be removed; but they've been useful.

mock/endpoints.ts
```diff
@@ -0,0 +1,25 @@
+export default {
```
portante: What is the significance of naming this file `mock/endpoints.ts`?  Is the `.ts` something special?

config/config.js
portante: Why are we not providing a "mock" of `/api/v1/endpoints` instead of allowing it to pass through?

portante (dismissed):

mock/endpoints.ts
dbutenhof: Answering both of your points in one, and I'd tried to capture this in my PR description but apparently failed... I found that UMI has a special place in its steely heart reserved for the `mock` directory, and processes any `*.ts` files early as global mocks. So the `endpoints.ts` _is_ the mock for the endpoints API, and it allows `document.ejs` (and the generated `index.html`) to unconditionally `fetch('/api/v1/endpoints')` in both mock and "real" environments; umi's mocking takes care of this transparently.

So this file creates a mock for anyone doing a `GET` on `/api/v1/endpoints`.

This works for me with Gurbir's `start:mock` environment, which pulls up a local dashboard page loaded entirely from local umi/npm dev server mocks. This file has no effect on `build` deployments, and my `fetch('/api/v1/endpoints')` works fine there, too.

The only problem is the e2e environment, which is doing something entirely different. Whether fixing that will require giving up on this neat trick remains to be seen. I'm hoping some of the npm/yarn/umi gurus will have some insight.

aquibbaig: I think `endpoints.js` works fine here since we don't static-type the object here (which can be an advantage of using a .ts file)

dbutenhof: Thanks, @aquibbaig; I'm mostly feeling my way through this blindly, as I have very little experience with the npm/yarn/umi development environment. When I discovered the references to the special `.ts` file type, I gained the mistaken impression that was special magic... and since it worked (at least for `start:mock` if not for `test:e2e`), I didn't question that assumption.

You're correct that it works exactly the same as `endpoints.js`. Not surprisingly, though, this also doesn't fix my e2e problems so it doesn't move me forward.

aquibbaig: @dbutenhof No problem! The e2e tests are fixed as a part of Gurbir's  react-table PR  now.

dbutenhof: Unfortunately, rebasing on Gurbir's change doesn't solve my e2e failures.

I changed `controllers.e2e.js` to specify `headless: false` and pulled up the debug tab as soon as the window came up. The root problem seems to (still) be that the initial `fetch ( '/api/v1/endpoints' )` is returning the entire `index.html` and apparently not hitting my mock at all.

For `start:mock` the endpoints `fetch` returns the mocked payload; but under the e2e environment it returns

```
<!DOCTYPE html><html lang="en"><head>

<link rel="stylesheet" href="/umi.css">

  <meta charset="UTF-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <title>Pbench Dashboard</title>
  <link rel="icon" href="/favicon.ico" type="image/x-icon">
  
  <!-- runtime endpoints config -->
  <script type="text/javascript">
    fetch( '/api/v1/endpoints' )
      .then(response => {
        console.log(`RESPONSE: ${response.status},${response.statusText}`);
        if (response.ok) {
          response.text().then(data => {
            console.log(`ENDPOINTS: ${data}`);
            window.endpoints = JSON.parse(data);
          }).catch(error => console.log(`ERROR data: ${error}`))
        } else {
          console.log(`NOTOK: ${response}`);
        }
      })
      .catch(error => console.log(`ERROR: ${error}`));
  </script>
<script>
  window.routerBase = "/dashboard/";
</script>
</head>

<body>
  <div id="root"></div>
<script src="/umi.js"></script>



</body></html>
```

And that breaks all of the selectors because the endpoint references like `endpoints.api.controllers_months` don't even resolve.

aquibbaig: @dbutenhof I just saw that the `controllers.e2e.js` file in your `endpoints` branch is referencing the older file and is different from the current https://github.com/distributed-system-analysis/pbench-dashboard/blob/main/src/e2e/controllers.e2e.js. Also, your endpoint branch looks to be at this state: `This branch is 3 commits ahead, 22 commits behind distributed-system-analysis:main.`

dbutenhof: Yes, after Gurbir's change went in I merged my local branch and tried things out, but without any change. I didn't bother pushing an update. Just for reference, I'll rebase again and push a new commit to get in sync.

mock/api.js
```diff
@@ -3,7 +3,7 @@ import casual from 'casual';
 
 const DEFAULT_SIZE = 100;
 
-export const mockIndices = new Array(DEFAULT_SIZE).fill().map(() => moment().format('YYYY-MM'));
+export const mockIndices = new Array(1).fill().map(() => moment().format('YYYY-MM'));
```
dbutenhof, line 6: Note, while there was nothing wrong with the previous version, it was generating an array of 100 duplicates of the exact same YYYY-MM value, which is useless.

Really, we should possibly consider randomly generating actual dates for the mocked datasets and collecting the months to represent here as if we were really querying the active run index month range....

mock/endpoints.ts
aquibbaig: Screenshot

@dbutenhof  Currently, `windows.endpoints` are getting referenced in the test environment too.

aquibbaig: I think what you're trying to do is use this endpoint to populate data in the mock environment, is that so?

aquibbaig: @dbutenhof here's what I found by testing this branch. Idk if it'd be useful or what you intend to do, so do let me know

We used React which listened for changes in the UI and updated the page dynamically, but that's not the case with HTML/ejs (unfortunately). That's why we do not see results unless we reload the page on manually script the data into some HTML tag using js. 

- What's happening here is when the mock environment starts and browser fires up, `umi.js` intercepts requests to `/controllers/list` here: https://github.com/distributed-system-analysis/pbench-dashboard/blob/3aae4358bd367b676832821c57f1152a4dfd7c46/src/e2e/controllers.e2e.js#L20.

- Then the dashboard populates the data from the mockfile. After the `umi.js` is called, our script tag is invoked, calling `/api/v1/endpoints`, which fetches the data and updates `window.endpoints`, but does not re-request data. But there is no hook to update the changes on the webpage. It just updates the endpoints and stays like that for the rest of the session. **Note:** umi.js is always invoked at the earliest. Some results from the Network tab.
```
- umi.js 
  -  queued at 48.29ms
  - started at 48.54ms
- endpoints script that we wrote
  - queued at 134.78ms
  - started at 135.40ms 
```
which means that the interception comes earlier than the custom script updating the endpoints, so the data is always fetched from the mock files in the dashboard and not the `/api/v1` endpoint.
 
-  Just manually reloading the browser (without hard reset/cache clean, just a normal reload) can be used to test if it works that way. We can use a `window.location.reload` here but I'd dig deeper into this to find the optimum solution, else there is a race condition between the e2e tests asserting values and fetch API updating the page.

aquibbaig:
@dbutenhof the tests are running fine on my local environment and I also am able to see window.endpoints being set correctly after `fetch` request. I cannot re-trigger **this** build unfortunately, please let me know if you can retrigger the build from your end

dbutenhof:
> @dbutenhof the tests are running fine on my local environment and I also am able to see window.endpoints being set correctly after `fetch` request. I cannot re-trigger **this** build unfortunately, please let me know if you can retrigger the build from your end

Really? They didn't (and don't) work for me locally, so I didn't even bother to push a new commit after my last rebase.

I can't restart the failed run, either; I suspect that's because they're shutting down the `travis-ci.org` site; the "Restart" button is grayed out and a popup says that the site is in "read-only mode".

I force-pushed the top of my branch, but I'm not sure that the Travis job will run at all. Webb's been working on backporting our new internal Jenkins-based unit test rig to the 0.69 branch, and it may be time for the dashboard to work out something similar, either Jenkins or GitHub Actions.

But I still don't understand why the tests aren't running for me locally. They used to be failing the same way as in the Travis runs, but I guess my local sandbox must be broken somehow if they're working for you now.

dbutenhof:
Oh, right; the dashboard *did* switch to GitHub Actions since I'd pushed the last commit on this branch to GitHub.

It's odd, because I don't see any indication of the status here; it just says "All checks have passed", referring to the LGTM run.

However I received an email that the GitHub Actions run had failed: https://github.com/dbutenhof/pbench-dashboard/runs/3230244073?check_suite_focus=true

The failure looks the same, but they're all just page load timeouts that don't help me at all to understand what might have happened. Does this help you?

aquibbaig:
> Oh, right; the dashboard _did_ switch to GitHub Actions since I'd pushed the last commit on this branch to GitHub.
> 
> It's odd, because I don't see any indication of the status here; it just says "All checks have passed", referring to the LGTM run.
> 
> However I received an email that the GitHub Actions run had failed: https://github.com/dbutenhof/pbench-dashboard/runs/3230244073?check_suite_focus=true
> 
> The failure looks the same, but they're all just page load timeouts that don't help me at all to understand what might have happened. Does this help you?

Thanks @dbutenhof. The build has now successfully passed

dbutenhof:
> Thanks @dbutenhof. The build has now successfully passed

The LGTM passed, but GitHub isn't showing a link to the GitHub Actions test run, and the log shows that the e2e tests are still failing. I tried re-running it, and it failed again.

I'm puzzled: where are you looking?

aquibbaig:
> > Thanks @dbutenhof. The build has now successfully passed
> 
> The LGTM passed, but GitHub isn't showing a link to the GitHub Actions test run, and the log shows that the e2e tests are still failing. I tried re-running it, and it failed again.
> 
> I'm puzzled: where are you looking?

@dbutenhof Hmm, so I was looking at my tests locally and they are passing, and I presumed from "All checks have passed" that Github Actions actually ran. It's weird that the Github Actions link is not here

aquibbaig:
But I can see on your branch that the tests have failed

aquibbaig:
There are a lot of page load timeouts in the test, which means that nothing is being rendered on the page (implying that there is some error when the dashboard loads in the mock environment), which is weird as it works fine locally.

dbutenhof:
> There are a lot of page load timeouts in the test, which means that nothing is being rendered on the page (implying that there is some error when the dashboard loads in the mock environment), which is weird as it works fine locally.

Yes, that's always been the problem. I agree with your analysis, and my conclusion (and Gurbir's) was that the generated `index.html` is being loaded before the UMI mocks are established.

I've run the e2e tests locally with `headless` disabled; using the Chrome console I can see that the `fetch('/api/v1/endpoints')` is returning the entire `index.html` instead of the endpoints mock, which breaks every page. (I always thought it was odd that the test framework engine was returning the entire `window.origin` page rather than an error getting the relative URI.)

We could never figure out **why** this was happening, or how to fix it. Gurbir's eventual conclusion was that we'd have to get rid of UMI and do our own packaging to make this work, but that was more a hope than a definite conclusion.

aquibbaig:
As the page loads, I can see that the data is logged, but JSON.parse fails?
Screenshot from 2021-08-04 16-05-48

Also, this is what I found in the trace:
Screenshot from 2021-08-04 16-05-53

@dbutenhof Can you try and see if this fetches the endpoints properly in the test environment.
```diff
export default {
-  'GET /endpoints': mockEndpoints,
+ 'GET /api/v1/endpoints': mockEndpoints,
};
```
in [mock files](https://github.com/dbutenhof/pbench-dashboard/blob/endpoints/mock/api.js)

dbutenhof:
Thanks for working on this, Aquib. I know I'd tried adding the `/api/v1` prefix before; just to be sure, I tried again this morning with the same result as always. I'd copied the style of the other mock endpoints in `api.js` so I presume it's doing either suffix or subset matching, and this shouldn't make a difference.

Note that the response that's dumped in the console shows the entire `index.html` contents, not the mocked endpoints JSON, which is why there's a JSON parsing error. It really looks like the `fetch` was processed without the mock in place, although I still don't understand why `fetch('/api/v1/endpoints')` returns the `window.origin` HTML instead of failing if nothing's been programmed to respond at that address. Maybe that's a fallback behavior of the testing environment?

So when we try to reference `window.endpoints.api` to make the call to `controllers/list`, there's no `api` key, because it's HTML rather than JSON. All of these observations are perfectly consistent with what I had seen before; I just have no idea at all how that's happening, or what to do next to solve the mystery. 😕 

This is a strange situation, and I clearly didn't explain it well enough initially; but it appears that you've reached the same brick wall I hit on your own. That's oddly comforting in that it's not "just me". 😄

aquibbaig:
@dbutenhof I had another dig at this today, I found that when I prefix `/api/v1` for `/endpoints` route in `mock/api.js`, all my test cases ran successfully. The above trace was only received when I didn't add the prefix. Can you please double check this on your end too? (I am testing it in the MOCK environment) My `mock/api.js` file export looks like this:
```
export default {
  'GET /api/v1/endpoints': mockEndpoints,
  'POST /datasets/list': (req, res) => {
    const data = {};
    data[req.body.controller] = mockResults(req.body.controller);
    res.send(data);
  },
  'POST /datasets/detail': (req, res) => {
    res.send(mockDetail(req.body.name));
  },
  'POST /datasets/toc': mockTableContents,
  'POST /datasets/samples': mockSamples,
  'POST /datasets/timeseries': mockTimeseries,
  'POST /mappings': mockMappings,
  'POST /search': mockSearch,
  'POST /sessions/list': mockSessions,
  'POST /sessions/create': mockSession,
};
```

Screenshot from 2021-08-11 16-41-51

dbutenhof:
> @dbutenhof I had another dig at this today, I found that when I prefix `/api/v1` for `/endpoints` route in `mock/api.js`, all my test cases ran successfully.

I've tried that locally several times, without any effect. However I've never tried pushing that up to GitHub, and I *have* seen results on the Travis/GitHub runs that differ from what I see locally (another annoyance I have never solved). I just tried again locally, but this time I'll try pushing a new commit to see what happens on the GitHub end. Maybe, in the end, the problem is just my local environment?

It's worth a try.

dbutenhof:
@aquibbaig I'm not sure why the GitHub Action wasn't running; but I don't see any GitHub Action activity on `pbench-dashboard` for over a week, and I suspect that the `push` trigger is only working when a PR is actually merged. When I'd experimented with GitHub Actions for pyesbulk, I'd used both `push` and `pull_request`. I tried adding that, and the Action ran for my new commit. I suspect this is what we should have on main.

In any case, the GitHub Action failed, but much less dramatically than I've seen repeatedly on my local system... so I think I must have a configuration issue that's interfering with the e2e tests.

However, it did still fail. Can you take a look at https://github.com/distributed-system-analysis/pbench-dashboard/runs/3300688029?check_suite_focus=true when you have a chance to see you have any idea about causes? Since I'm not seeing this on my local system (and apparently neither are you) this is likely to be difficult to debug.

aquibbaig:
> @aquibbaig I'm not sure why the GitHub Action wasn't running; but I don't see any GitHub Action activity on `pbench-dashboard` for over a week, and I suspect that the `push` trigger is only working when a PR is actually merged. When I'd experimented with GitHub Actions for pyesbulk, I'd used both `push` and `pull_request`. I tried adding that, and the Action ran for my new commit. I suspect this is what we should have on main.
> 
> In any case, the GitHub Action failed, but much less dramatically than I've seen repeatedly on my local system... so I think I must have a configuration issue that's interfering with the e2e tests.
> 
> However, it did still fail. Can you take a look at https://github.com/distributed-system-analysis/pbench-dashboard/runs/3300688029?check_suite_focus=true when you have a chance to see you have any idea about causes? Since I'm not seeing this on my local system (and apparently neither are you) this is likely to be difficult to debug.

I just re-ran the tests (there were some tests where "sorting" fails in the table)

dbutenhof:
Yes, most were labeled "sorting" issues; but in the tests with any actual diagnostics, I saw messages indicating that it was expecting 20 or 50 outputs but got only 1. I wonder what could have caused this?

Even more concerning ... I clicked the re-run button in the GitHub Action, and the tests have now passed. That may mean the errors are due to some sort of instability in the test scripts, possibly due to timing; which means we can't completely depend on the automated testing.

I'm going to squash this PR into one commit, since the history probably isn't very useful at this point, and then mark it "Ready for review". More importantly, we'll see whether the new test run succeeds or fails... and how.

And then, at some point, I'm going to have to try to figure out why my local e2e tests aren't working at all because it's apparently just me. Fun, fun.

dbutenhof:
Thanks to @aquibbaig this PR is now running e2e tests cleanly. (Most of the time: there seems to be some non-reproducible timing element in the tests that causes occasional failures.)

I've also deployed this build to my server VM and it works. I guess it's ready (finally) for review.

src/services/dashboard.js
```diff
@@ -76,13 +75,11 @@ export async function queryTocResult(params) {
...
     endpoints,
-    endpoints.run_toc_index,
+    endpoints.indices.run_toc_index,
     selectedDateRange
   )}/_search`;
 
-  const endpoint = MOCK_UI
-    ? `${endpoints.pbench_server}/datasets/toc`
-    : `${endpoints.pbench_server}/elasticsearch`;
+  const endpoint = MOCK_UI ? endpoints.api.datasets_toc : endpoints.api.elasticsearch;
```
webbnh, line 92: Are these two endpoints really swappable?  And, why does the code need to know whether its dependency is being mocked?...can't we go with just one choice and mock that when we want to test?  (Ditto for lines 104 and 190, et al.)

.github/workflows/main.yml
```diff
@@ -1,5 +1,5 @@
 name: CI
-on: [push]
+on: [pull_request, push]
```
webbnh, line 2: When Gurbir initially switched the CI to GitHub Actions, I ended up doing some research on it, concerned about it being (ab)used as an attack vector.  I believe that my conclusion was that just `push` was correct...but, I no longer recall why.  🙁  (It's possible that I wasn't taking into account forked repos at that time.)  I hope this does what we want it to (and nothing else...).

src/utils/moment_constants.js
```diff
@@ -28,10 +28,10 @@ export function getAllMonthsWithinRange(endpoints, index, selectedDateRange) {
...
   }
   monthResults.forEach(monthValue => {
-    if (index === endpoints.result_index) {
-      queryString += `${endpoints.prefix + index + monthValue}-*,`;
+    if (index === endpoints.indices.result_index) {
+      queryString += `${index + monthValue}-*,`;
     } else {
-      queryString += `${endpoints.prefix + index + monthValue},`;
+      queryString += `${index + monthValue},`;
     }
```
webbnh, line 35: This would be DRYer:
```
    queryString += `${index + monthValue}`;
    if (index === endpoints.indices.result_index) {
      queryString += `-*`;
    }
    queryString += `,`;
```

webbnh (commented):
I'm not much of a front-end person, but, since you asked me to look, below are a few comments.

I'm not really sure how the conditionals in the CUT to support mocking are really helping anything...it seems like there ought to be a better approach to that which is more transparent for the main code.

.github/workflows/main.yml
dbutenhof: I'd done research before that when I set up GitHub Actions on the pyesbulk repo, and at the time decided that I needed both, though I no longer recall the details of why; understanding precisely the conditions and context of the webhook events proves extraordinarily complex because each segment of the documentation is fairly shallow and doesn't provide the vertical context I want.

I would have expected that `push` would track pushing a commit to a PR branch; however, it clearly does not, and  `pbench-dashboard` seems to be currently running the Action only when a PR is merged into `main`, which isn't what we want.

The [documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) suggests, in a fairly convoluted and inconclusive way, that `pull_request_target` may be less of a potential exposure, but basically goes on to point out exactly the same potential issues as `pull_request` and I don't see any more appropriate target.

src/services/dashboard.js
dbutenhof: The mocks used to trigger on data within the Elasticsearch JSON payload, because the actual URIs are identical.

Gurbir switched to this model because it means each "API" has a distinct URI that can be reliably and straightforwardly mocked. Once we've replace the direct Elasticsearch calls with Pbench APIs, these splits will go a way. Right now, the production code needs to make the elasticsearch calls, but using the planned Pbench API URIs vastly simplifies the mocks.

src/utils/moment_constants.js
dbutenhof: Yes, maybe, although building the string through three separate steps isn't as readable. This code will go away once we can drop the remaining direct Elasticsearch queries, anyway.

dbutenhof (commented):
> I'm not much of a front-end person, but, since you asked me to look, below are a few comments.

I put you on the list because of your renowned attention to detail but mostly just to make you aware of it. 😬 

> I'm not really sure how the conditionals in the CUT to support mocking are really helping anything...it seems like there ought to be a better approach to that which is more transparent for the main code.

At the time, Gurbir thought that was a less ugly hack than the way the mocks worked before, because they had to pick apart the JSON query payload since they weren't distinct URIs. It was ugly and difficult to maintain. Aside from the cruft in the production code, which certainly is unfortunate, this makes the test infrastructure more maintainable and that's a good thing.

The alternatives will evaporate once we finish the Pbench server query APIs, so it's a temporary hack... then the production front end code will have distinct URIs for each query and the mocking will be straightforward.

src/services/dashboard.js
webbnh: > Right now, the production code needs to make the elasticsearch calls

I figured that that was the case, but that's the part which is confusing:  don't the two endpoints require _different_ payloads?  (Or, is there another analogous conditional in the code which constructs the payloads, too?)

.github/workflows/main.yml
webbnh: > I would have expected that push would track pushing a commit to a PR branch

Without brushing up on the documentation, I would expect the same but only if it were pushed to the same repo and not to a fork of it....

mock/api.js
```diff
@@ -281,21 +287,59 @@ export const mockSession = {
   },
 };
 
+export const mockEndpoints = {
```
aquibbaig, line 290: Hey @dbutenhof, what are your thoughts on whether we should keep `/api/v1` prefix for all the endpoints here since when this corresponds to mock? So, we can do something like /api/mock/controllers/list and change the corresponding handlers too

aquibbaig (dismissed):
Looks good to me

mock/api.js
dbutenhof: So far as I can tell it doesn't make much difference as the mock mechanism seems to trigger based on a substring match; however, since you had to add the prefix to get the `fetch` from `index.html` to trigger, I thought it looked odd to have the others without the prefix so I added them everywhere. And that means the `get` or `post` from the dashboard code needs to include the prefix or it won't match ... so this makes the `mockEndpoints` JSON consistent with the `api.js` mocks.

The problem with changing to use a mocked path like `/api/mock` would be in that `document.ejs` (`index.html`) `fetch`; I think we'd need to go back to using a config variable to make EJS change the path based on whether we're building for test or production, or we'll break the `fetch('/api/v1/endpoints')` call again. I really liked the idea of making the URI a literal so there's no difference in the generated page.

src/services/dashboard.js
dbutenhof: Only one payload, although that payload will change once we shift over to the backend API. Right now, the dashboard test environment isn't trying to pretend that `/api/v1/datasets/toc` is returning the fully post-processed data we anticipate for the Pbench API ... it's just the raw Elasticsearch output the dashboard effect currently expects. We're just faking the URI to make it easier to mock. Yeah, it's weird, but the old mocking code that was based on searches of the JSON query payload was even more weird. 😬 

And, again, it's only until we actually implement those APIs ...

mock/api.js
```diff
@@ -281,21 +287,59 @@ export const mockSession = {
...
+    controllers_months: '/api/v1/controllers/months',
+    datasets_list: '/api/v1/datasets/list',
+    datasets_detail: '/api/v1/datasets/detail',
+
+    // NOTE: these are not implemented in the server yet, but are defined
+    // below as mocks for testing
+    datasets_toc: '/api/v1/datasets/toc',
+    datasets_samples: '/api/v1/datasets/samples',
+    datasets_timeseries: '/api/v1/datasets/timeseries',
+    mappings: '/api/v1/index/mappings',
```
npalaska, line 302: I guess it should be `/api/v1/index/mappings/` since that how its implemented in server now

webbnh (approved):

config/config.js
portante: No longer a relevant comment.

mock/api.js
dbutenhof: You should add that when you change the dashboard to actually use it. 😁

portante (approved):

PR #143, opened 2021-04-08
render breadcrumbs on page

### LINKED ISSUE #128: render breadcrumbs on pageheader component
To improve navigation, we might want to add breadcrumbs to every page. The way to go here is to make use of Pattenfly's [breadcrumb component](https://www.patternfly.org/2020.06/documentation/react/components/breadcrumb).

aquibbaig:
Fixes #128


anishaswain:
@aquibbaig is this PR ready for review yet?

aquibbaig:
@anishaswain Actually I am waiting for Mary for some designs regarding the placement of breadcrumbs w.r.t header mast. We're almost there implementation wise

anishaswain:
@aquibbaig @gurbirkalsi while going through the PR, I noticed that two words like `Comparison Select` or `Run Comparison` looks a bit funky on the breadcrumbs. Will it be a good idea to combine both words with a `-` or CamelCase instead?
Screenshot 2021-05-05 at 4 35 27 PM

aquibbaig:
image
*went ahead with the second design

src/common/menu.js
```diff
@@ -37,19 +37,35 @@ const menuData = [
...
+    // For all non root paths,
+    // append the parent name with
+    // current name.
+    if (level !== 1) {
+      name = `${parentName}/${name}`;
+    }
+    let result = {};
     if (item.routes) {
-      result.routes = formatter(item.routes, `${parentPath}${item.path}`, item.authority);
+      result = {
```
gurbirkalsi: We can avoid duplicate assignment to `result` in the if else block by moving it outside of the block (DRY principle).

src/components/LoginHint/index.js
```diff
@@ -1,7 +1,9 @@
 import React, { Component } from 'react';
 import { routerRedux } from 'dva/router';
 import { connect } from 'dva';
-import { Title, Button } from '@patternfly/react-core';
+import { Title, Button, Banner, Flex, FlexItem } from '@patternfly/react-core';
```
gurbirkalsi, line 4: These changes seem unrelated to the PR. Did you mean to include changes to update the login hint UI/UX?

src/layouts/BasicLayout.js
```diff
@@ -87,27 +82,52 @@ class BasicLayout extends React.PureComponent {
...
-    return {
-      location,
-      breadcrumbNameMap: this.breadcrumbNameMap,
-      routes: route.routes,
-    };
+  componentDidMount() {
+    const {
+      location: { pathname },
+    } = this.props;
+    let exactPath = {};
```
gurbirkalsi: We should consider creating a function for the logic on lines 94 - 102 to avoid repeating duplicate functionality (DRY principle).

src/utils/utils.js
```diff
@@ -104,3 +104,38 @@ export const isLoggedInUser = username => {
   }
   return false;
 };
+
+export const isRootPath = pathname => {
+  if (pathname === '/' || pathname === '/search' || pathname === '/sessions') {
```
gurbirkalsi: We should consider importing the `menuData` object in `menu.js` instead of hardcoding the root path entrypoints for the dashboard. We can do this by making `menuData` exportable and iterating over each object in the array.

src/components/LoginHint/index.js
aquibbaig: Yeah, so we had a discussion in the design sync to add the login banner rather for a hint

src/components/Breadcrumb/index.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    return [...pathTo(route.parent), route];
+  };
+
+  const navigateToCrumb = path => {
+    dispatch(routerRedux.push(path));
+  };
+
+  return (
```
anishaswain, line 20: @aquibbaig trying using the below code for rendering arrow icon.
```
<Breadcrumb>
      {pathTo(context).map((crumb, index, breadcrumbs) => (
            <BreadcrumbItem
            className={index < breadcrumbs.length - 1?styles.active:''}
            key={crumb.path}
            onClick={() => navigateToCrumb(crumb.path)}
            isActive={index === breadcrumbs.length - 1?true:false}
          >
            {crumb.name}
          </BreadcrumbItem>
        )
      )}
</Breadcrumb>
 ```

**Review by anishaswain — CHANGES_REQUESTED** [review]

File: src/components/Breadcrumb/index.js (hunk 5 above)
Comment (aquibbaig) [reply src/components/Breadcrumb/index.js:20]: That works, thanks @anishaswain

**Comment by aquibbaig** [comment]
@anishaswain mergeable?

**Review by anishaswain — COMMENTED** [review]
@aquibbaig LGTM! @aquibbaig could you please squash the commits before we merge.

File: src/common/menu.js [hunk 6]
```diff
@@ -37,23 +37,37 @@ const menuData = [
...
+      ...(withParent.routes && {
+        routes: setupParents(withParent.routes, withParent),
+      }),
+    };
+  });
+
+const flattenRoutes = routes =>
+  routes.map(route => [route.routes ? flattenRoutes(route.routes) : [], route]).flat(Infinity);
+
+const getMenuData = () => flattenRoutes(setupParents(buildPaths(menuData)));
```
portante, line 71: Why do we still need this?

portante (changes requested):

src/common/menu.js
aquibbaig: Good question @portante!
umijs requires you to define your routing configuration in a config file. We do not instantiate any top-level declaration w.r.t routes as we would have done with something like  [react-router-dom](https://reactrouter.com/web/guides/quick-start).
So, we generate the breadcrumbs array (required by the patternfly breadcrumbs component) by doing some post-processing based on menudata (used for routing) and the current location of the user (passed in by the page). Earlier there was a lot of unnecessary post-processing which I have removed in this PR.

portante (approved):

anishaswain (dismissed):
LGTM! Just need to resolve the conflicts. @aquibbaig could you please resolve the conflicts.

portante (changes requested):
@aquibbaig, looks like this PR needs a rebase.

portante (dismissed):

PR #148, opened 2021-04-28
Pass logged-in user to controller query

dbutenhof:
Pass the logged-in user state through from the controller page to the service and on to the back end.

This appears to work on my server VM, although no controllers will be reported as the datasets are currently all owned by "pbench" and with the dashboard's insistence on valid email addresses there's no way to create a user that the server will see as "pbench". Cie la vie. For the jam session we'll upload datasets owned by actual users.

### DIFF
--- a/src/pages/Controllers/index.js
+++ b/src/pages/Controllers/index.js
@@ -69,10 +69,10 @@ class Controllers extends Component {
   };
 
   fetchControllers = () => {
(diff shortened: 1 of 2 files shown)


gurbirkalsi (approved):

aquibbaig (approved):
LGTM

PR #149, opened 2021-04-29
enable onRow click property for react table

aquibbaig:
We used to make the record name bold on hover which shifted the table columns to the right by some pixel value. 

What this PR does:
- [x] Highlights the row selection on a hover.
- [x] Enable row click property by a handler function passed as props to Table component.
- [x] Stop default propagation of rowclick for row actions such as selecting checkboxes and favoriting records.

### DIFF
--- a/src/components/Table/index.js
+++ b/src/components/Table/index.js
@@ -62,7 +62,7 @@ function fuzzyTextFilterFn(rows, id, filterValue) {
 
 fuzzyTextFilterFn.autoRemove = val => !val;
 
-function Table({ columns, data, isCheckable, onCompare, loadingData }) {
+function Table({ columns, data, isCheckable, onCompare, loadingData, onRowClick }) {
   const filterTypes = React.useMemo(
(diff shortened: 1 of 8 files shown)


aquibbaig:
TODO:
If we disable font-weight change on hover, then there is no visual cue other than row highlight as implemented above. I went ahead with the implementation and added a `rowClick` property to the Table, but now I am not able to focus on favoriting records or selecting a run comparison checkbox, so I will investigate on fixing that issue.

aquibbaig:
> TODO:
> If we disable font-weight change on hover, then there is no visual cue other than row highlight as implemented above. I went ahead with the implementation and added a `rowClick` property to the Table, but now I am not able to focus on favoriting records or selecting a run comparison checkbox, so I will investigate on fixing that issue.

Should be fixed now

aquibbaig:
Closing due to the fact that this is a PR in upstream (by mistake) and will be now tracked under #153

PR #152, opened 2021-05-05
Remove remaining ant design dependencies

gurbirkalsi:
### Summary
Remove remaining ant design dependencies except for the Date Picker component. ant design automatically handles tree shaking for components not used with library imports which helps minimize bundle size.

Remove unused node modules from package.json and upgrade umi.js to v3 for handling css/less/img imports with new PatternFly dependency versions.


aquibbaig:
@gurbirkalsi Can we not use `@` imports after the customization of `config/config.js`, the tests are failing due to that

anishaswain (commented):
LGTM! 
Just a couple of queries.
1. Are we continuing to use ant design for `datepicker` component?
2. I think the tests are failing because they can't find the modules used with `@`. Was it a feature/plugin of ant design which is being removed?

portante (changes requested):
Just needs a successful set of test runs.

gurbirkalsi:
> LGTM!
> Just a couple of queries.
> 
> 1. Are we continuing to use ant design for `datepicker` component?

Yes, we'll keep ant-design in our dependency tree to support the DatePicker component. ant-design automatically handles tree shaking to remove any components we don't use as part of our final application bundle which lets us reduce bundle size.

> 2. I think the tests are failing because they can't find the modules used with `@`. Was it a feature/plugin of ant design which is being removed?

It seems the upgrade to umi.js v3 is causing this issue. Thanks for catching this - I'll debug and have a fix up shortly!

portante (approved):

anishaswain (approved):

PR #153, opened 2021-05-05
enable onRow click property for react table

aquibbaig:
We used to make the record name bold on hover which shifted the table columns to the right by some pixel value. 

What this PR does:
- [x] Highlights the row selection on a hover.
- [x] Enable row click property by a handler function passed as props to Table component.
- [x] Stop default propagation of rowclick for row actions such as selecting checkboxes and favoriting records.

### DIFF
--- a/src/components/Table/index.js
+++ b/src/components/Table/index.js
@@ -62,7 +62,7 @@ function fuzzyTextFilterFn(rows, id, filterValue) {
 
 fuzzyTextFilterFn.autoRemove = val => !val;
 
-function Table({ columns, data, isCheckable, onCompare, loadingData }) {
+function Table({ columns, data, isCheckable, onCompare, loadingData, onRowClick }) {
   const filterTypes = React.useMemo(
     () => ({
       fuzzyText: fuzzyTextFilterFn,
@@ -139,7 +139,7 @@ function Table({ columns, data, isCheckable, onCompare, loadingData }) {
                 </div>
               ),
               Cell: ({ row }) => (
-                <div>
+                <div onClick={e => e.stopPropagation()}>
                   <IndeterminateCheckbox {...row.getToggleRowSelectedProps()} />
                 </div>
               ),
@@ -202,7 +202,11 @@ function Table({ columns, data, isCheckable, onCompare, loadingData }) {
               {page.map(row => {
                 prepareRow(row);
                 return (
-                  <tr className="pf-m-hoverable" {...row.getRowProps()}>
+                  <tr
+                    className="pf-m-hoverable"
+                    {...row.getRowProps()}
+                    onClick={onRowClick && onRowClick(row.original)}
+                  >
                     {row.cells.map(cell => {
                       return (
                         <td {...cell.getCellProps()}>
--- a/src/components/Table/index.less
+++ b/src/components/Table/index.less
@@ -1,8 +1,18 @@
 tr > td > span > a {
-  text-decoration: none !important;
   color: inherit !important;
+  text-decoration: none !important;
 }
 
 .table-wrapper {
   overflow-x: auto;
+}
+
+a:hover {
+  font-weight: normal !important;
+}
+
+tr:is(.pf-m-hoverable):hover {
+  background: var(--pf-global--BackgroundColor--light-200);
+  border-left: 4px solid var(--pf-c-card--m-selectable--m-selected--before--BackgroundColor);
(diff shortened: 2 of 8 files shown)


src/components/Table/index.less
```diff
@@ -1,8 +1,7 @@
...
-  color: inherit !important;
-}
-
 .table-wrapper {
   overflow-x: auto;
 }
+
+tr:is(.pf-m-hoverable):hover {
+  background: aliceblue;
+}
```
anishaswain, line 9: @aquibbaig I am still a little reluctant about adding a complete different colour on the hover action. Instead how about using PatternFly's provided default value `var(--pf-c-table--tr--m-hoverable--active--BackgroundColor)` ?

Also, I think it would be great to make a little bit more customisation to make it look like  PF table. How about recreating the [style of odd rows](https://www.patternfly.org/2020.04/documentation/react/components/table#custom-row-wrapper) for active/ hoverable rows?

aquibbaig: @anishaswain Good point regarding reusing the Patternfly color variable, let me add that to the code. Also, can you share some designs for the suggested customization? The above link shows a 404 for me

anishaswain: @aquibbaig I have updated the link

anishaswain:
@aquibbaig how about this design?
Screenshot 2021-05-05 at 4 26 31 PM

The CSS for the same would be 
```
background: var(--pf-global--BackgroundColor--light-200);
border-left: 4px solid var(--pf-c-card--m-selectable--m-selected--before--BackgroundColor);
box-shadow: var(--pf-c-table--tbody--m-hoverable--hover--BoxShadow);
```

src/components/Table/index.less
gurbirkalsi: Just a heads up @anishaswain - the link you shared seems to point to an old release of PatternFly 3. We'll want to make sure we reference documentation from PatternFly 4's latest release.

aquibbaig:
> @aquibbaig how about this design?
> Screenshot 2021-05-05 at 4 26 31 PM
> 
> The CSS for the same would be
> 
> ```
> background: var(--pf-global--BackgroundColor--light-200);
> border-left: 4px solid var(--pf-c-card--m-selectable--m-selected--before--BackgroundColor);
> box-shadow: var(--pf-c-table--tbody--m-hoverable--hover--BoxShadow);
> ```

I didn't understand why `any other row` is highlighted here @anishaswain

anishaswain:
> I didn't understand why `any other row` is highlighted here @anishaswain

Ahh, my bad. That's just because I added css for one row in console and I happened to hover on another row as well while taking screenshot. The motive was to show the design of the row.  With the implementation, only one row will be highlighted on hover.

src/components/Table/index.less
anishaswain: @gurbirkalsi the link points to `2020.04` documentation. The link is `https://www.patternfly.org/2020.04/documentation/react/components/table#custom-row-wrapper`. Let me know if I missed something here.

gurbirkalsi (changes requested):
Clicking on a controller in the "Controllers" page causes the "Results" page to have empty data in the mock environment. We'll need to revisit and make sure we're passing the correct data after the on click event.

gurbirkalsi:
Looks like Anisha is tackling this problem with PR #163 - thanks @anishaswain!

aquibbaig:
Thanks @gurbirkalsi

src/components/Table/index.less
gurbirkalsi: The latest release should be `2021.08` at [https://www.patternfly.org/v4/components/table#custom-row-wrapper](url).

src/components/Table/index.js
```diff
@@ -202,7 +202,11 @@ function Table({ columns, data, isCheckable, onCompare, loadingData }) {
               {page.map(row => {
                 prepareRow(row);
                 return (
-                  <tr className="pf-m-hoverable" {...row.getRowProps()}>
+                  <tr
+                    className="pf-m-hoverable"
+                    {...row.getRowProps()}
+                    onClick={onRowClick ? () => onRowClick(row.original) : null}
```
gurbirkalsi: What if we did  `onClick={onRowClick && () => onRowClick(row.original)}` for a single condition?

gurbirkalsi (approved):
Small point to consider, otherwise looks good!

PR #154, opened 2021-05-06
add error handling logic and patternfly alerts

### LINKED ISSUE #147: Integrate new error handling into API requests
User authentication PR #130 intoduces a new error handling mechanism to show Pf alerts on errors and success. We need to integrate that logic to other API requests as well.

aquibbaig:
What this PR does

- Adds API error handling logic to the redux store
- Adds patternfly alerts to page components to show error and success messages.


portante (changes requested):
The error handling pattern seems replicated in each instance.  Is there a way to encapsulate that to avoid the need to replicate it?

aquibbaig:
> The error handling pattern seems replicated in each instance. Is there a way to encapsulate that to avoid the need to replicate it?

@portante 
We can use a common message from a separate file and invoke it here. For the error handling pattern, I believe each pattern needs to be customized here (it is not expected to be the same everywhere). Each fallback, whenever there is an error from API request, should be different for different requests. For example: For date range picker, if `/api/aliases` throws an error we fall back to default dates of "last week", that's how the DatePicker component does not throw an error and the dashboard keeps running.

- A custom response from the server on error:
Response -> { data: message: "xyz" } which is a possible reason due to python's `abort()`.
- We take this message from the response and display it on the dashboard, if there is no such message we use a custom message as implemented. For more clarity, I believe this custom message can be invoked from a separate file and should be mapped from the status code. (for example: custom message for 404 should be "not found", and for "500" should be "internal server error")

aquibbaig:
will track this is another PR, a lot of stuff implemented here is unnecessary and repetitive

aquibbaig:
tracked by #185

PR #161, opened 2021-05-19
Phase out private routes from routing schema

### LINKED ISSUE #145: Change routing structure with auth in the dashboard
We need to implement higher level component wrappers to handle auth verification before pushing the user to a "protected" route. The goal is to phase out the private route so that authentication is only handled within the javascript bundle and is not exposed via routes.

gurbirkalsi:
Pbench dashboard will verify user auth status according to successful or unsuccessful API responses given a valid user token. In cases where the server returns an error, the user will be routed to authentication pages.


portante (approved):

portante (changes requested):
Well, it looks good to me, but for some reason the unit tests are not passing.

portante (approved):

aquibbaig (approved):

PR #163, opened 2021-06-01
Fixes mock API for replicating server and avoid post-processing

anishaswain:
- Fixes mock API for replicating server(now mock will send data exactly in the format, expected from server) and avoid post-processing of data.

- Fixes rendering of results page in mocked environment

### DIFF
--- a/mock/api.js
+++ b/mock/api.js
@@ -6,41 +6,24 @@ const DEFAULT_SIZE = 100;
 export const mockIndices = new Array(DEFAULT_SIZE).fill().map(() => moment().format('YYYY-MM'));
 
 export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index) => ({
-  controller: casual.word + index,
-  key: casual.word + index,
-  last_modified_string: moment.utc() + index,
+  controller: `${casual.word}.${casual.word}.com`,
+  key: casual.ip,
+  last_modified_string: moment().valueOf(moment.utc()),
   last_modified_value: moment.utc() + index,
   results: casual.integer(1, DEFAULT_SIZE),
 }));
 
-export const mockResults = {
-  took: DEFAULT_SIZE,
-  hits: {
-    total: {
-      value: DEFAULT_SIZE,
-    },
-    hits: new Array(DEFAULT_SIZE).fill().map((value, index) => ({
-      _index: moment().format('YYYY-MM'),
-      _type: casual.word,
-      _id: casual.uuid,
-      _score: null,
-      _source: {
-        '@metadata': {
-          controller_dir: casual.word,
-        },
-        run: {
-          controller: casual.word + index,
-          name: casual.word,
-          start: moment.utc() + index,
-          end: moment.utc() + index,
-          id: casual.uuid,
-          config: casual.word,
-        },
-      },
-      sort: [casual.unix_time],
-    })),
-  },
-};
+export const mockResults = hostname =>
+  new Array(DEFAULT_SIZE).fill().map(() => ({
+    '@metadata.controller_dir': hostname,
+    config: casual.word,
+    controller: hostname,
+    end: moment.utc(),
+    // Since dataset id is a long hex string, removed "-" characters here to make it look like real data
+    id: casual.uuid.replace(/-/g, ''),
+    result: casual.word,
+    start: moment.utc(),
+  }));
 
 export const mockSamples = {
   _scroll_id: casual.uuid,
@@ -150,42 +133,46 @@ export const mockSamples = {
   },
 };
 
-export const mockDetail = {
-  hits: {
-    hits: [
-      {
-        _source: {
-          run: {
-            id: 'test_id',
-            controller: 'test_controller',
-            name: 'test_result',
-            script: casual.random_element(['fio', 'uperf']),
-            date: moment.utc(),
-            start: moment.utc(),
-            end: moment.utc(),
-            user: casual.username,
-          },
-          host_tools_info: [
-            {
-              hostname: 'test_host',
-              tools: {
-                iostat: '--interval=3',
-                mpstat: '--interval=3',
-                perf: "--record-opts='record -a --freq=100'",
-                pidstat: '--interval=30',
-                'proc-interrupts': '--interval=3',
-                'proc-vmstat': '--interval=3',
-                sar: '--interval=3',
-                turbostat: '--interval=3',
-              },
-            },
-          ],
-        },
-      },
-    ],
+export const mockDetail = hostname => ({
+  hostTools: new Array(DEFAULT_SIZE).fill().map(() => ({
+    hostname: hostname.split('.')[0],
+    'hostname-f': hostname,
+    tools: {
+      'hostname-alias': '',
+      'hostname-all-fqdns': hostname,
+      'hostname-all-ip-addresses': casual.ip,
+      'hostname-domain': `${casual.word}.${casual.word}.com`,
+      'hostname-fqdn': hostname,
+      'hostname-ip-address': casual.ip,
+      'hostname-nis': 'hostname: Local domain name not set',
+      'hostname-short': hostname.split('.')[0],
+      'rpm-version': `v0.+${casual.integer}`,
+      tools: casual.word,
+      vmstat: '',
+    },
+  })),
+  runMetadata: {
+    config: casual.word,
+    controller: hostname,
+    controller_dir: 'gprfc056.sbu.lab.eng.bos.redhat.com',
+    date: moment.utc(),
+    end: moment.utc(),
+    start: moment.utc(),
+    'file-date': moment.utc(),
+    'file-name': `/${casual.work}/${casual.work}/${casual.work}`,
+    'file-size': casual.building_number,
+    id: casual.uuid.replace(/-/g, ''),
+    iterations: casual.word,
+    md5: casual.uuid.replace(/-/g, ''),
+    name: casual.word,
+    'pbench-agent-version': `v0.+${casual.integer}`,
+    raw_size: casual.building_number,
+    script: casual.word,
+    'tar-ball-creation-timestamp': moment.utc,
+    'toc-prefix': casual.word,
+    toolsgroup: 'default',
   },
-};
-
+});
 export const mockTableContents = {
   hits: {
     total: DEFAULT_SIZE,
@@ -297,8 +284,14 @@ export const mockSession = {
 export default {
   'GET /controllers/months': mockIndices,
   'POST /controllers/list': mockControllers,
-  'POST /datasets/list': mockResults,
-  'POST /datasets/detail': mockDetail,
+  'POST /datasets/list': (req, res) => {
+    const data = {};
+    data[req.body.controller] = mockResults(req.body.controller);
+    res.send(data);
+  },
+  'POST /datasets/detail': (req, res) => {
+    res.send(mockDetail(req.body.name));
+  },
   'POST /datasets/toc': mockTableContents,
   'POST /datasets/samples': mockSamples,
   'POST /mappings': mockMappings,


mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
-          id: casual.uuid,
-          config: casual.word,
-        },
-      },
-      sort: [casual.unix_time],
-    })),
-  },
-};
+export const mockResults = new Array(DEFAULT_SIZE).fill().map((value, index) => ({
+  '@metadata.controller_dir': casual.word,
```
dbutenhof: Normally, I would expect `@metadata.controller_dir` to match `controller`. I'm not sure that the current dashboard code cares, but the results of the `/datasets/list` query are by definition all for the same "controller" as reported by `/controllers/list` (which reports controllers owned by a specified user within a specified time period).

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  key: casual.uuid,
```
dbutenhof: The original dashboard effect didn't have `key`, however looking at the Chrome console shows complaints about mising key in arrays all over the place. I don't think I added key into my implementation of `/datasets/list` ... but I probably should have. However, I think it should be the same value as `result` (the name of the dataset).

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  id: casual.uuid,
```
dbutenhof: Hmm. `casual.uuid` returns standard UUID 8-4-4-4-12 format, whereas the dataset `id` is the Elasticsearch document ID, which is the MD5 hash of the tarball... a long hex string like `3646953262b0adfe6d1c32903912087b`. Again, I don't think the dashboard code cares what it represents, so this is a nit; but mocking data that looks something like the real data is a better strategy. I don't think that `casual` has the generator we really want here, and UUID isn't likely to break anything. Maybe all it needs is a comment, or maybe removing the `-` characters.

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  config: casual.word,
+  controller: casual.word + index,
```
dbutenhof: `controller` is usually the hostname where the benchmark command was run: using `casual.url` might be better. (Note that the example I found shows just name.domain, although the word "url" implies something like http://name.domain, and the latter wouldn't be appropriate here... a numeric `casual.ip` might work instead.)

And the `controller` will be the same for all datasets returned by the query.

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  deletion: moment.utc().add('days', 30),
```
dbutenhof: Several of your fields are actually requests for changes. Neither `deletion`, `seen`, `saved`, nor `status` has any precedent. This is fine, but best to defer these for a later PR. We need to add them on the server side, and the mocks should be mocking what the server returns!

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  end: moment.utc() + index,
```
dbutenhof: The real query response has both `start` and `end` ... they're timestamps, so simply adding `index` to the string result isn't a good idea. And you're missing `start`...

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  result: casual.word,
+  seen: false,
```
dbutenhof: We need to define how this idea is represented and manipulated on the server side. Maybe it's just a Dataset metadata that the dashboard can set. I don't have any objection to adding the field to the `/datasets/list` query, although I do have an objection to adding it to the dashboard mocks now; let's try to keep the mocks in sync with the server, or we're not testing that the dashboard code can work with the server.

mock/api.js
```diff
@@ -13,34 +13,19 @@ export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index)
...
+  saved: Math.random() < 0.5,
+  status: 'published',
```
dbutenhof: What other values do you propose for `status`?? In the Elasticsearch documents, I've defined `owner` and `access` fields, where access will start out as `private` and can be changed to `public` by publishing it. I chose to represent it that way to support eventual expansion to groups of users, such that the value of `access` might in the future be a group name defining a set of users who can access it.

Along the same lines, we might have `access` here with a direct pass-through of the document access. Or we could keep it simple and for now just use `published: true | false`. In any case, again, I don't think we should be expanding the mock in this PR; just fix it. We should discuss how we want to change the query going forward, and then define issues for both the dashboard and server to implement.

mock/api.js
```diff
@@ -297,7 +282,11 @@ export const mockSession = {
 export default {
   'GET /controllers/months': mockIndices,
   'POST /controllers/list': mockControllers,
-  'POST /datasets/list': mockResults,
+  'POST /datasets/list': (req, res) => {
+    const data = {};
+    data[req.body.controller] = mockResults;
+    res.send(data);
+  },
```
dbutenhof: This is interesting, because you're taking the `req.body.controller`, the controller name we're requesting (as returned by the `/controllers/list` mock), but returning the list of datasets with randomly named controllers. Really, each element in the mock array should be that same `req.body.controller`. I wonder if it would be better to generate the mock here rather than trying to re-use a static mock value? You're generating the outer JSON dynamically anyway.

Also, really all of the mocks could be done this way so the response matches the request, but that wasn't done with the others. We ought to be consistent. I like the idea of the response matching; it's just odd to do it only here.

dbutenhof (changes requested):

mock/api.js
gurbirkalsi: +1, at a minimum the dashboard should care about the datatypes it is ultimately rendering. In this case, we want to make sure we are processing a string for `@metadata.controller_dir`. In order to make our mock tests easier to debug, we should name mocked fields according to the schema we've come up with in Elasticsearch.

mock/api.js
aquibbaig: +1

mock/api.js
anishaswain: Fixed to match the `@metadata.controller_dir` to `controller` name. Also `controller` is in `name.name.com` format to replicate `https:\\name.domain now`. Also using `casual.ip` was causing issue with e2e test.

mock/api.js
anishaswain: Replaced '-' characters in uuid to make it look like real data

mock/api.js
anishaswain: resolved

mock/api.js
anishaswain: Fixed

mock/api.js
anishaswain: I am removing the context of this new `access` or `status` field from this PR. These will be referenced in #105

mock/api.js
anishaswain: Fixed the controllers to match. Had to update other queries too.

mock/api.js
anishaswain: I have removed these new fields for now

mock/api.js
anishaswain: Removed these. Will add these filed in overview page PR #105 
(these fields will be needed there for data rendering)

anishaswain:
@dbutenhof  What does the value represent in `"last_modified_value": 1621954963008.0` in query `POST /api/v1/controllers/list`

mock/api.js
```diff
@@ -6,41 +6,24 @@ const DEFAULT_SIZE = 100;
 export const mockIndices = new Array(DEFAULT_SIZE).fill().map(() => moment().format('YYYY-MM'));
 
 export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index) => ({
-  controller: casual.word + index,
-  key: casual.word + index,
-  last_modified_string: moment.utc() + index,
+  controller: `${casual.word}.${casual.word}.com`,
+  key: casual.ip,
+  last_modified_string: moment.utc(),
   last_modified_value: moment.utc() + index,
```
dbutenhof: These are supplied by Elasticsearch as part of the aggregation. The query aggregates results by the `"run.start"` field. It's a date datatype, and Elasticsearch apparently stores it internally as a number; presumably a UNIX seconds (or milliseconds) since the Epoch, though I never bothered to work that out. It returns both that binary internal representation and the string date. Both were passed up from the model before, and I left it that way ... although I'm not at all sure the dashboard code actually cares.

This sort of thing would be good to consider when we think about the additions you want to make; we don't have to be tied to the "raw" Elasticsearch data where that doesn't help us.

mock/api.js
```diff
@@ -6,41 +6,24 @@ const DEFAULT_SIZE = 100;
 export const mockIndices = new Array(DEFAULT_SIZE).fill().map(() => moment().format('YYYY-MM'));
 
 export const mockControllers = new Array(DEFAULT_SIZE).fill().map((value, index) => ({
-  controller: casual.word + index,
-  key: casual.word + index,
-  last_modified_string: moment.utc() + index,
+  controller: `${casual.word}.${casual.word}.com`,
```
dbutenhof, line 9: That's an interesting way to do it.

I didn't think to mention this before; however to be really consistent the value returned here should match the `hostname` parameter in your `/datasets/list` mock; because this is the controller name the dashboard will be requesting there.

If the dashboard cared about that consistency, this would be an issue.

mock/api.js
```diff
@@ -6,41 +6,24 @@ const DEFAULT_SIZE = 100;
...
+  key: casual.ip,
```
dbutenhof, line 10: I don't think we use `key` for anything, but React complains if arrays don't have a "unique key"; so this is probably just fine. However the real code (the current server code as well as the original dashboard model) will set this to the controller name.

mock/api.js
```diff
@@ -151,40 +134,60 @@ export const mockSamples = {
...
-                sar: '--interval=3',
-                turbostat: '--interval=3',
-              },
-            },
-          ],
-        },
-      },
-    ],
-  },
+  hostTools: new Array(DEFAULT_SIZE).fill().map(() => ({
```
dbutenhof, line 137: There are two keys in the JSON doc returned by `/datasets/detail`: `"hostTools"` and `"runMetadata"`.  You're only returning the `hostTools`.

```
  "runMetadata": {
    "config": "test-03",
    "controller": "gprfc056.sbu.lab.eng.bos.redhat.com",
    "controller_dir": "gprfc056.sbu.lab.eng.bos.redhat.com",
    "date": "2021-05-24T03:27:21",
    "end": "2021-05-24T03:28:29.949026",
    "file-date": "2021-05-24T03:28:25.292000",
    "file-name": "/srv/pbench/archive/fs-version-001/gprfc056.sbu.lab.eng.bos.redhat.com/pbench-user-benchmark_test-03_2021.05.24T03.27.21.tar.xz",
    "file-size": 5944316,
    "id": "3646953262b0adfe6d1c32903912087b",
    "iterations": "1-default",
    "md5": "3646953262b0adfe6d1c32903912087b",
    "name": "pbench-user-benchmark_test-03_2021.05.24T03.27.21",
    "pbench-agent-version": "v0.71.0-185g5527f23f1",
    "raw_size": 6828923,
    "script": "pbench-user-benchmark",
    "start": "2021-05-24T03:27:55.061895",
    "tar-ball-creation-timestamp": "2021-05-24T03:28:38.255317",
    "toc-prefix": "pbench-user-benchmark_test-03_2021.05.24T03.27.21",
    "toolsgroup": "default"
  }
```

dbutenhof (changes requested):

mock/api.js
webbnh: According to [EpochConverter](https://www.epochconverter.com/) (which is my go-to tool for this stuff), that's milliseconds since the UNIX Epoch, which is Midnight, 1 January 1970 UTC.

It probably should be an integer (although, fractional milliseconds _do_ exist ;-) ), but apparently it somehow got converted to a float somewhere in the many layers between there and here.

dbutenhof: I see Elasticsearch returning the timestamp in normalized exponential form: `"value": 1.617290619102E12`. The Pbench `/controllers/list` API just copies the value.

webbnh: I remember chasing all of this stuff down in a previous life.  ES has its own foibles, and then that's compounded by the requirements of JSON during transport, and then it's all subject to the fact that the JSON gets converted to a Python 'dict' with Python datatypes...so, who knows?!

mock/api.js
anishaswain: Fixed !

mock/api.js
anishaswain: fixed!

mock/api.js
anishaswain: Fixes it to use milliseconds since the UNIX Epoch

mock/api.js
```diff
@@ -297,8 +284,16 @@ export const mockSession = {
 export default {
   'GET /controllers/months': mockIndices,
   'POST /controllers/list': mockControllers,
-  'POST /datasets/list': mockResults,
-  'POST /datasets/detail': mockDetail,
+  'POST /datasets/list': (req, res) => {
+    const data = {};
+    data[req.body.controller] = mockResults(req.body.controller);
+    console.log(data);
```
aquibbaig: we can remove logs from this PR

aquibbaig (approved):
LGTM

mock/api.js
portante: Yes, please remove.

portante (dismissed):
Can you squash to one commit after removing the `console.log()` call?

mock/api.js
```diff
@@ -6,41 +6,24 @@ const DEFAULT_SIZE = 100;
...
-  },
-};
+export const mockResults = hostname =>
+  new Array(DEFAULT_SIZE).fill().map(() => ({
+    '@metadata.controller_dir': hostname,
+    config: casual.word,
+    controller: hostname,
+    end: moment.utc(),
+    // Since dataset id is a long hex string, removed "-" characters here to make it look like real data
+    id: casual.uuid.replaceAll('-', ''),
```
gurbirkalsi: Seeing the following error in development environment:

```
id: _casual.default.uuid.replaceAll('-', ''),
                           ^

TypeError: _casual.default.uuid.replaceAll is not a function
```

dbutenhof: Older browser version? The `replaceAll` is apparently new. Instead, `casual.uuid.replace(/-/g, '');` ... or use something other than `uuid` here, since what we want is really a "long hex string" (pretending to be an MD5 hash of the document).

mock/api.js
dbutenhof: I'd prefer that the mocks try to model the form and semantics of the real APIs; it may not matter now, where the dashboard code doesn't care, but it might matter in the future and it's likely easier to get it right now than to make someone figure out what's going on in a year when the processing changes. :-)

Unfortunately, that means converting this into a full lambda method with local variables so that a value can be used more than once.

mock/api.js
```diff
@@ -150,42 +133,46 @@ export const mockSamples = {
...
+    config: casual.word,
+    controller: hostname,
+    controller_dir: 'gprfc056.sbu.lab.eng.bos.redhat.com',
+    date: moment.utc(),
+    end: moment.utc(),
+    start: moment.utc(),
+    'file-date': moment.utc(),
+    'file-name': `/${casual.work}/${casual.work}/${casual.work}`,
+    'file-size': casual.building_number,
+    id: casual.uuid.replaceAll('-', ''),
```
dbutenhof: I assume some test environments will run into the same old browser version Gurbir hit with `replaceAll`? Should they all be changed "to be safe"???

dbutenhof (approved):
Although Gurbir's comments warn that some developers may try running this against older browser versions without `replaceAll`, it doesn't seem to be an issue for the Travis environment, and I don't know if we have a firm policy in place for browser version support. I'm going to approve, although I have some concerns about this issue.

aquibbaig (commented):
Any comments on this error trace I am getting on results page?

```
/pbench-dashboard/mock/api.js:32
  id: _casual.default.uuid.replaceAll('-', ''),
                           ^

TypeError: _casual.default.uuid.replaceAll is not a function

```

webbnh:
@aquibbaig, Gurbir noted the [same problem](https://github.com/distributed-system-analysis/pbench-dashboard/pull/163#discussion_r657637849) -- did you see Dave's responses ([1](https://github.com/distributed-system-analysis/pbench-dashboard/pull/163#discussion_r659713314), [2](https://github.com/distributed-system-analysis/pbench-dashboard/pull/163#discussion_r659715504))?

mock/api.js
anishaswain: removed !

mock/api.js
anishaswain: replaced with `casual.uuid.replace(/-/g, '');`

mock/api.js
anishaswain: I have replaced it with `casual.uuid.replace(/-/g, '');`. Hopefully it will work now.

mock/api.js
```diff
@@ -297,8 +284,15 @@ export const mockSession = {
...
   'POST /controllers/list': mockControllers,
-  'POST /datasets/list': mockResults,
-  'POST /datasets/detail': mockDetail,
+  'POST /datasets/list': (req, res) => {
+    const data = {};
+    data[req.body.controller] = mockResults(req.body.controller);
+    res.send(data);
+  },
+  'POST /datasets/detail': (req, res) => {
+    console.log(req.body);
```
aquibbaig: we can remove these logs

mock/api.js
```diff
@@ -297,8 +284,15 @@ export const mockSession = {
 export default {
   'GET /controllers/months': mockIndices,
   'POST /controllers/list': mockControllers,
-  'POST /datasets/list': mockResults,
-  'POST /datasets/detail': mockDetail,
+  'POST /datasets/list': (req, res) => {
+    const data = {};
+    data[req.body.controller] = mockResults(req.body.controller);
+    res.send(data);
```
aquibbaig, line 290: as we are reassigning to a const, we can do something like let data = {} or return {req.body.controller: mockResults(req.body.controller)}

aquibbaig (changes requested):
Awesome work @anishaswain, thanks for fixing the bug on the `Results` page which was "blocking" my work. Just some minor changes and we are good to go 🚀

mock/api.js
anishaswain: @aquibbaig I believe we can't send the response object like that. It's throwing error.
Screenshot 2021-07-01 at 6 02 05 PM

anishaswain: We can't use `let` either as data isn't getting reassigned. 
Screenshot 2021-07-01 at 6 05 01 PM

mock/api.js
anishaswain: removed!

aquibbaig:
> @aquibbaig, Gurbir noted the [same problem](https://github.com/distributed-system-analysis/pbench-dashboard/pull/163#discussion_r657637849) -- did you see Dave's responses ([1](https://github.com/distributed-system-analysis/pbench-dashboard/pull/163#discussion_r659713314), [2](https://github.com/distributed-system-analysis/pbench-dashboard/pull/163#discussion_r659715504))?

Thanks @webbnh, they have been fixed in Anisha's latest commit

mock/api.js
anishaswain: @aquibbaig let me know if you have any suggestions for further modification here.

aquibbaig (approved):
Looks good to me!

mock/api.js
dbutenhof: I'm not going to push on this issue further in this PR, though I still worry a bit that eventually we'll run into problems from the inconsistency of mocks. (For example, that the controller names in this response are randomly generated words, while we're using a `hostname` variable to configure the controller names in `mockResults`.

It may be that the advantages of pushing the dashboard code with our randomized data outweighs the cost of trying to generate data that follows the expected server schema; and it's probably not worth a lot of effort right now.

mock/api.js
dbutenhof: The `const` is fine because the JSON document itself is only created once; what's weird about the `= {req.body.controller: ...}` form is that you're trying to assign multiple levels of JSON at once, which doesn't work. It'd have to be `{req: {body: {controller: ...}}}` instead. Which may or may not be easier to read.

However, I think that what you're actually doing here is assigning the literal key `"controller"` in your JSON, whereas the real value is going to be the name of the specific controller.

PR #164, opened 2021-06-01
fixes feature of closing a selected filter

anishaswain:
https://user-images.githubusercontent.com/16955978/120318252-43816500-c2fd-11eb-8f29-7addfa5b1f33.mov

### DIFF
--- a/src/components/Select/index.js
+++ b/src/components/Select/index.js
@@ -5,11 +5,18 @@ import {
   SelectOption,
   SelectVariant,
   SelectDirection,
+  ChipGroup,
+  Chip,
 } from '@patternfly/react-core';
 
 export default class Select extends PureComponent {
   static propTypes = {
     options: PropTypes.array.isRequired,
+    selected: PropTypes.array,
+  };
+
+  static defaultProps = {
+    selected: [],
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/components/Select/index.js
```diff
@@ -54,12 +56,34 @@ export default class Select extends PureComponent {
     }
   };
 
+  chipGroupComponent = selected => {
+    const { onSelect, category } = this.props;
+    return (
+      <ChipGroup>
+        {(selected || []).map(currentChip => {
```
gurbirkalsi: When `selected` is empty, using .map() will not return `<Chip>` tags. We can omit the conditional for `(selected || [])` and use `selected.map()`. For further confidence in type verification, we can use PropTypes validation to ensure `selected` is an empty array if a value for it is not passed during component references.

gurbirkalsi (changes requested):

aquibbaig (approved):

src/components/Select/index.js
anishaswain: @gurbirkalsi I have resolved the same and have added default value of `selected` to `PropTypes`

gurbirkalsi (approved):

PR #166, opened 2021-06-07
filters table while searching(autocomplete)

### LINKED ISSUE #157: Table search autocomplete
During a discussion with @mnmehta, we identified a requirement for autocompleting search results when searching within the table component for allowing faster navigation.

aquibbaig:
This PR fixes #157

There may be some instances where `onSearch` is null/undefined, depending on whether we want to activate the autocomplete feature, for those cases, we can provide a custom prop `autocomplete: Boolean` to the Table instantiation and conditionally activate `onSearch` property. To test these cases, we have to fix our mock API or staging server to reproduce Table data. (currently, these are empty arrays in every other page other than controllers, so I could only test on the controllers page)

### DIFF
--- a/src/components/SearchBar/index.js
+++ b/src/components/SearchBar/index.js
@@ -24,7 +24,12 @@ export default class SearchBar extends PureComponent {
   }
 
   onChange = value => {
+    const { onSearch } = this.props;
     this.setState({ searchValue: value });
+
+    // filters the Table as
+    // the input component changes
+    onSearch(value);
   };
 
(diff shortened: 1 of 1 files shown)


gurbirkalsi (approved):

portante (approved):

aquibbaig:
@gurbirkalsi mergeable?

gurbirkalsi:
If you can send a blank commit & rebase to trigger the new CI, we can confirm tests & merge!

PR #167, opened 2021-06-09
Deploy updates

portante:
Updates the deployment on the `b2.1` branch to include auto-tagging and copying of the SHA1 remotely so that we can verify what was deployed.

See also https://docs.npmjs.com/cli/v7/commands/npm-version#description and the description about how to use the `preversion`, `version`, and `postversion` scripts.


dbutenhof (commented):
I'm not familiar with the npm `version` stuff, and I don't know where that came from since it's not on `main` ... but it looks plausible enough. The Ansible changes look straightforward. Mostly, it seems that all of this should probably be on `main`, too, since none of this looks specific to the old ES1 dashboard.

package.json
```diff
@@ -1,13 +1,19 @@
 {
   "name": "pbench_dashboard",
-  "version": "2.1.0",
+  "version": "2.1.1",
```
aquibbaig, line 3: I am not sure but the version in my deployed dashboard shows `2.1.1-0`

aquibbaig (approved):
Looks good!

gurbirkalsi (approved):

anishaswain (approved):
LGTM!

dbutenhof (approved):

package.json
portante: How did you deploy it?

portante:
I'll file another PR to forward port these changes to the `main` branch.

package.json
aquibbaig: @portante I deployed it on localhost, changing some parts of main.yml

PR #169, opened 2021-06-16
Utilize GitHub Actions for CI jobs in place of Travis CI

gurbirkalsi:


webbnh:
@gurbirkalsi, did you see that the check failed?

Also, is there a vector here through which we could be abused?  That is, if someone were to send a PR with a change to this file, could they get GitHub to execute arbitrary actions on their behalf, without us accepting the change?

gurbirkalsi:
Yes! Seems to be an issue with waiting for the port to become available after starting the development environment. 

Yes, potentially - we can ensure owners of the repo must review changes to the GitHub actions config by creating a "CODEOWNERS" file. I will include that file with this PR. 

https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners

webbnh:
Are the various tracebacks in the unit tests and end-to-end tests expected behavior?

gurbirkalsi:
Not exactly; however, we should address those warnings in a separate PR. Most of them are relatively harmless and won't impact dashboard functionality. Ideally the testing output should be free of warnings in addition to errors.

aquibbaig (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Run Unit Tests
+        uses: borales/actions-yarn@v2.0.0
+        with:
+          cmd: test # will run `yarn test` command
+      - name: Run E2E Tests
+        shell: bash
+        run: |
+           yarn start:mock &
+           sleep 2m &&
+           curl http://localhost:8000 &&
```
portante: Instead of sleeping for 2 minutes, could we do something like: https://stackoverflow.com/questions/11904772/how-to-create-a-loop-in-bash-that-is-waiting-for-a-webserver-to-respond/50583452?

portante: This is the kind of thing we do in the `e2etests` script.  Is there a way to just invoke that `e2etests` script directly?

portante (changes requested):

.github/workflows/main.yml
gurbirkalsi: This worked! Much cleaner, thanks @portante.

webbnh: @gurbirkalsi, you have a somewhat generous definition of "worked".  😁 

I love the simplicity of the `main.yml` file, now, but I'm concerned about merging this in while it's producing a failed result.

webbnh: OK, so it passed on its most recent run...I'm not sure what changed (I don't even see what triggered the re-run of the checks).  Are the results somehow non-deterministic?

For posterity, here's the failure which has subsequently disappeared:
```
Running E2E tests ...
yarn run v1.22.10
$ umi test ./src/e2e --verbose
FAIL src/e2e/controllers.e2e.js (13.325 s)
  controller page component
    ✓ should load controllers (5124 ms)
    ✓ should search for controller name (84 ms)
    ✓ should reset search results (66 ms)
    ✕ should sort controllers column alphabetically ascending (93 ms)
    ✓ should sort controllers column alphabetically descending (67 ms)
    ✓ should sort last modified column chronologically ascending (85 ms)
    ✓ should sort last modified column chronologically descending (71 ms)
    ✓ should sort results column numerically ascending (67 ms)
    ✓ should sort results column numerically descending (63 ms)
    ✓ should display 20 controllers in the table (53 ms)
    ✓ should display 50 controllers on preference (87 ms)
    ✓ should display 75 controllers on preference (126 ms)
    ✓ should display 100 controllers on preference (129 ms)
    ✓ should display the date picker component on click (112 ms)
    ✓ should change the selected dates after picking them from the calendar (287 ms)

  ● controller page component › should sort controllers column alphabetically ascending

    expect(received).toBe(expected) // Object.is equality

    Expected: "accusamus41"
    Received: "accusamus6"

      100 |       elem => elem.innerHTML
      101 |     );
    > 102 |     expect(testController).toBe(
          |                            ^
      103 |       mockControllers.sort((a, b) => (a.controller > b.controller ? 1 : -1))[0].controller
      104 |     );
      105 |   });

      at Object.<anonymous> (src/e2e/controllers.e2e.js:102:28)
```

webbnh: I realize that the contents (and stability) of the tests are probably out-of-scope for this PR, but I regard testing as foundational, so I'm quite uncomfortable with the notion that the tests are "passing" while issuing warnings or that they sometimes spontaneously fail.  I haven't done a lot of front-end work, so perhaps I'm just naive and that this is just the way it is...but I hope to heck that that is not the case!

So, I have two issues:  
- first, I'd love to get some reassurance the test problems that we're seeing here really do have nothing to do with the fact that they are being run under GitHub Actions (meaning that they should not gate this PR);
- second, do we have (or should someone like me) open issues against the dashboard for these warnings?  (@anishaswain, @aquibbaig, @dbutenhof, @portante)

gurbirkalsi: The previous test run failed due to duplicate data generated from the mock environment. This is an issue @aquibbaig opened here: https://github.com/distributed-system-analysis/pbench-dashboard/issues/136

Ideally, the tests expect data from our mock environment that is easily assertable. This can be improved by either running postprocessing on generated mock datasets to remove duplicates, selecting data from a "predefined" list of items, or adding unique IDs to each generated value. All approaches would allow us to avoid issues of duplication & ensure consistent tests going forward.

This should be addressed in a separate PR that includes changes to the mocked environment. For the scope of this PR, we are only moving CI functionality from TravisCI to GitHub Actions.

webbnh:
> we can ensure owners of the repo must review changes to the GitHub actions config by creating a "CODEOWNERS" file. I will include that file with this PR.

It turns out that this doesn't actually address the problems with execution of malicious code:  this change would enforce review by the core engineers, but the testing is run _before_ that happens.  So, as far as I'm concerned, you can omit this file from this PR -- I don't think it does anything that we need.

The capabilities of the GitHub Actions are limited by the settings of the repo.  So, to try to minimize the possibilities for malfeasance, I've changed the settings to grant only read access to actions and to limit actions to running only local actions or actions defined by GitHub.  In this way, if someone pushes a change to the `main.yml` file which causes it to run some malicious script (also included in the pushed branch), or if they just push a change to add something nasty to the unit test script, it won't have write access to the repository...I think.

.github/workflows/main.yml
```diff
@@ -0,0 +1,14 @@
+name: CI
+on: [push]
```
npalaska, line 2: Shall we also add `pull_request` as an event here? Or just `push` enough?

.github/workflows/main.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  build:
+    name: Test
+    runs-on: ubuntu-latest
```
npalaska, line 6: Its interesting that GitHub-hosted runner only has ubuntu in linux [github hosted runner](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on) but thats okay I think because we used debian with travis as well

npalaska (commented):
Looks good

.github/workflows/main.yml
webbnh: Since you cannot do a pull request without having first done a push, I _think_ (for these purposes, anyway) adding `pull_request` would be redundant.

.github/workflows/main.yml
portante: In the future, we can change this to use `podman` and run our tests in a container of the kind we need.

portante (approved):

PR #172, opened 2021-06-22
Integrate dynamic routing via UmiJS

### LINKED ISSUE #99: Need of routing changes to incorporate addressable URLs
Every URL entered directly by the user in the browser, needs to be addressable by both logged in and non-logged-in users. 
The dashboard needs to handle the URL in such a manner that, it will either invoke a particular change in data(or, display of certain elements) according to the existence of the user. Or, it will simply redirect the user to the log in screen.

### LINKED ISSUE #100: URLs to specific records
Currently, we have a specific routing scheme as Controllers -> Results -> Summary. What happens if I would want to view the results of a controller, without going from the Controllers page. It would be as simple as appending a controller name or id to the url, or something of that sort to support "GET" requests to that URL.

### LINKED ISSUE #133: Consider implementing URLs for pages to allow sharing via bookmarks
Today, the dashboard has a notion of a "session" where the Redux state can be saved to a GraphQL instance under a URL that can be shared.  This requires saved state in some database that has to be managed.

Instead, we should consider the principle of not implementing behaviors that are already solved by well accepted / known methods.

If we encode the state of a page as a URL in the browser, and removed the need to save the state in a GraphQL instance, it would allow us to:

 * Leverage existing URL bookmarking mechanisms in the users existing browser
 * Leverage existing URL shortening services that exist today
 * Avoid having to maintain code and infrastructure for saved URLs
 * Avoid issues with logged-in vs not-logged-in users and how to save such state

gurbirkalsi:
This integration allows page components to handle reading from local props and window URL to parse data payloads. Additionally, this removes the notion of a "private" route and introduces the "overview" naming schema for routes to provide insight into a user's incoming runs to pbench.


portante (dismissed):

gurbirkalsi:
@aquibbaig @anishaswain keep in mind these changes do not include dynamic routing for the run comparison page due to the size of data payloads to reproduce the page. We'll need to come up with a clever way of rebuilding the page using either a compressed version of the original payload or only rebuild a subset of the run comparison page for dynamic links.

aquibbaig:
I think appending to the URL is a bad idea

anishaswain (dismissed):
LGTM for now. But we need to think about how to compress large payloads for result comparison.

aquibbaig:
@anishaswain what would be the state of run comparisons for now? (what route will they be in)

anishaswain (commented):
Having unexpected errors after rebase. Working on it
Screenshot 2021-07-19 at 4 03 45 PM

pravins:
We are not using UmiJS anymore in the new dashboard.

PR #183, opened 2021-07-19
pbench-server endpoint should use window.origin to access mock data

anishaswain:


aquibbaig (approved):

dbutenhof (approved):
Sure; since the mocks trigger on the URL suffix, the `/api/v1` doesn't matter for testing. It seems like an empty string would work just as well. (What _is_ `window.origin` in unit tests and e2e?)

PR #185, opened 2021-07-26
Enable error handling for API requests

### LINKED ISSUE #147: Integrate new error handling into API requests

### LINKED ISSUE #187: page unresponsive when error is thrown
The page becomes unresponsive and the network requests do not go through once an error is thrown (expected behaviour).
But we should not let users be blocked once errors are thrown, so we need to catch these errors.

Steps to reproduce this problem: (in b2.1 prod for example)
- Go to Explore page
- Head back to the controllers page and click on a controller, it will not make a network request to fetch a controller and the table shows "no data".

aquibbaig:
- adds error handling as patternfly toast alerts, fixes #187
- modify error model to add error and success messages
- modify request package to emit error message to the store
- error handling for auth model
- error handling for dashboard and datastore
- error handling for search and sessions
- rename auth services
- mocks user login, tests alert component


src/models/auth.js
```diff
@@ -19,9 +17,9 @@ export default {
     },
     /* eslint-disable camelcase */
     *loginUser({ payload }, { call, put, all }) {
-      try {
-        const response = yield call(queryLoginUser, payload);
-        const { auth_token, username } = response;
+      const { response, data } = yield call(postLoginUser, payload);
+      if (response.ok) {
+        const { auth_token, username } = data;
```
npalaska, line 22: Here you seem to be only handling when the response is ok, but what happens when response is not ok?

src/models/auth.js
```diff
@@ -32,133 +30,65 @@ export default {
...
-          payload: errortext,
-        });
       }
     },
     *registerUser({ payload }, { call, put, all }) {
-      try {
-        const response = yield call(queryRegisterUser, payload);
-        const { message } = response;
+      const { response, data } = yield call(postRegisterUser, payload);
+      if (response.ok) {
```
npalaska, line 37: Same question as above

npalaska (commented):
Looks good to me but I do have a question regarding how we are handling errors when the response is not `ok` on a specific query.

src/models/auth.js
aquibbaig: Good question @npalaska, and I forgot why I hadn't handled that. Ah, yes. the not ok cases are those which are made by `abort(status_code)` in the backend, these raise an exception on the dashboard, which directly is handled by the custom error handler [defined here](https://github.com/distributed-system-analysis/pbench-dashboard/blob/d2be3493ea1201c255f5be9f075c5b7c00dfcc21/src/utils/request.js#L6)

npalaska:
Looks good, but needs rebase I guess

npalaska (dismissed):

portante (commented):
I have reviewed this change set, but found it difficult to do given lots of indentation changes.

For the query/post name changes, in the future, having those in a separate commit would help the reviewer handle the volume of changes in pieces.

Further, where `if (response.ok) {` was added, might have been better to use the negative, `if (!response.ok) { return; }` instead, since that would prevent re-indenting code.

Do we have test cases for all the new methods added in `error.js`?

aquibbaig:
@portante Thanks for reviewing this PR, I agree on the point that I should have made individual commits which I did but I squashed them at the last moment before the merge (my bad!). I will go ahead and add test cases as you stated, thing is that it is hard to produce errors in the mock environment. I have an idea on how to do that though,

- I will add a mock user with username `admin` and pwd `admin`.
- When entered the wrong details it throws an error which should be caught by our procedure.
- clearing that error would close the alert

I think we can test it this way.

portante (approved):

npalaska (approved):

PR #191, opened 2021-08-12
Adds a unique key generator to utils for mock API

### LINKED ISSUE #178: inconsistent results for e2e tests
I noticed that casual.js generated namespaces sometimes can have the same key for different objects. That way, the Table component we have will be unable to sort the entries. We get inconsistency in terms of our tests, as a result, I have to re-rerun job multiple times, we should get a permanent fix for it (else dump the tests for sorting ascending or descending, keeping in mind that this function is already tested in react-table and should always work as expected)

anishaswain:
fixes #178 

As we are using result / controller name as key in the tables, we have to create unique names for the same, every-time in mock API. The `createUniqueKey` function helps in achieving that.


mock/api.js
```diff
@@ -21,7 +22,7 @@ export const mockResults = hostname =>
     end: moment.utc(),
     // Since dataset id is a long hex string, removed "-" characters here to make it look like real data
     id: casual.uuid.replace(/-/g, ''),
-    result: casual.word,
+    result: createUniqueKey(),
```
aquibbaig: can we do something like `casual.word + createUniqueKey` to make it sound a bit like a controller name rather than a hash-like string? I think this will just be for readability purposes

dbutenhof: I think this is the difficult case; `mockControllers` is a static payload that's generated *once* when `api.js` is loaded, whereas `mockResults` is a function that's run each time there's a call to `/datasets/detail`, meaning that the list of datasets reported for a given controller will change each time we make the call.

Also, we're now using your `createUniqueKey()` for the controller names reported by `/controllers/list` while here we're using `hostname` for the controller. (Line 21.)

That is, we may report a set of controllers for the date range including controller `1234567890abcdef` in `/controllers/list` and then when we call `/datasets/list` with that controller name, we're given a list of datasets with `createUniqueKey` "result" names which claim to have been generated from controller "hostname" instead of `1234567890abcdef`. As I said in the sync meeting, I don't think this impacts the behavior of the summary view page code, but it definitely isn't an accurate "mock" of the real user experience. I worry that this may complicate understanding and testing in the future, even though I can't currently construct any sequence where it's a real problem at present.

An alternative might be to generate a set of `mockResults` for each randomly generated controller name, using that controller name and with a random "result" name, at the same time `mockControllers` is called when we load `api.js`, perhaps in a JSON document using the controller name as key. Then, the "dynamic mock" for `/results/list` would return `mockResults[controller]` and we'd have consistent data across multiple calls. (At the expense of more complicated up-front processing.)

Another possibility would be for the `datasets/list` mock to dynamically generate a `mockResults` payload, but to *cache* the generated value in `mockResults[controller]` and re-use it if called again with the same controller rather than generating a new payload.

dbutenhof: Similarly, `mockDetail` passes the dataset *name* and uses a randomly generated controller. If these could be connected and cached, the UI "flow" would be more consistent. It probably doesn't matter to the automated tests at this point, but when running with the `start:mock` development server it'd more closely correspond to the real user experience. It's hard to weigh how much this really matters.

dbutenhof: I think this may also provide another approach to ensuring uniqueness of the keys. You could, for example, generate names in sequence: `controller1`, `controller2` and so forth; and `dataset1`, `dataset2`, and so forth.

src/utils/uniqueKey_generator.js
```diff
@@ -0,0 +1,12 @@
+export default function createUniqueKey() {
+  let dt = new Date().getTime();
+  const tempStr = 'xxxxxx.yxyxxyxy.xxxyyy';
+  const uuid = tempStr.replace(/[xy]/g, function(c) {
+    /* eslint no-bitwise: 0 */
+    const r = (dt + Math.random() * 16) % 16 | 0;
+    dt = Math.floor(dt / 16);
+    /* eslint no-bitwise: 0 */
+    return (c === 'x' ? r : (r & 0x3) | 0x8).toString(16);
+  });
```
webbnh: Does line 7 actually accomplish anything?  (It looks like its result is never used.). At line 6, what is the effect of adding `Math.random() * 16` if you're going to do a modulo 16 operation on the sum?  Isn't the result the same as `dt % 16`??  Were you intending to do the division before the modulo?

What does the `| 0` at line 6 do?  (I'm sure it does something wonderfully clever...I just can't remember what...does it give you an unsigned result?  Is it ensuring that you have a 32-bit result?)

Since you're already doing bitwise operations (and since 16 is a magic number), I recommend replacing the modulo operator in line 6 with a bitmask:
```
    const r = (dt + Math.random() * 16) & 0xf | 0;
```

Regarding the ESLint directives, it looks like you have [better options](https://eslint.org/docs/rules/no-bitwise) than just disabling it.
```
    /*eslint no-bitwise: ["error", { "int32Hint": true }] */
    const r = (dt + Math.random() * 16) % 16 | 0;
    /*eslint no-bitwise: ["error", { "allow": ["&", "|"] }] */
    return (c === 'x' ? r : (r & 0x3) | 0x8).toString(16);
```

webbnh: Also, this code is triggering an LGTM warning, which I think you could consider suppressing, by adding this to the end of line 6:
```
  // lgtm[js/insecure-randomness]
```

portante (changes requested):
I share the same concerns as the previous comments already made, but would like to add another consideration on how to approach this work: build a mock dataset once with multiple entries with names of the mock'd sets so that tests can request data be in place to help ensure certain test cases are covered.

mock/api.js
anishaswain: @dbutenhof to maintain the consistency across the `mockResults` I have used the `hostname`, which is the result of `mockControllers` (so we can safely assume that they are unique), and added a static value(the index of the results) to them for creating result names. In that way every time we call `api.js` with the same controller name, we will get the same result names. 

However, I agree that caching technique  would be more appropriate for the use cases like `overview-page` but I would like to address that in another PR and change the models to accommodate `cache`

src/utils/uniqueKey_generator.js
anishaswain: Removed line 7 (left it by mistake), Added correct ESLint directives.

mock/api.js
```diff
@@ -1,27 +1,28 @@
...
-  new Array(DEFAULT_SIZE).fill().map(() => ({
+  new Array(DEFAULT_SIZE).fill().map((value, index) => ({
     '@metadata.controller_dir': hostname,
     config: casual.word,
     controller: hostname,
     end: moment.utc(),
     // Since dataset id is a long hex string, removed "-" characters here to make it look like real data
     id: casual.uuid.replace(/-/g, ''),
-    result: casual.word,
+    result: `${index}${hostname.slice(0, -6)}${index}`,
```
dbutenhof, line 25: This makes the data a bit more consistent; however `mockDetail` is still randomly generating the dataset name even though it's reflecting the requested controller/host name now. It would be good to fix that, as well.

I'm not really happy with the overall structure of the dashboard API mocks we've managed to accumulate, and it seems that each small step forward in stability and consistency has been a big step "sideways" in that we're really not reflecting the expected schema of the actual data in many ways. The dashboard tests are currently mostly insensitive to this, but I worry it's going to be an increasingly serious problem as the dashboard gets smarter.

Still, we're agile, and a small step forward remains a small step forward. With some residual reluctance, I'll approve this and just hope we can take more time to improve this further in the future.

dbutenhof (approved):

portante (changes requested):
Seems worth while to try to move this forward as well, but the tests are failing.

portante:
Closed in favor of #202.

PR #192, opened 2021-08-12
Fixes inconsistent behaviour of "favourite controller" feature

### LINKED ISSUE #188: Inconsistent behaviour of "favourite controller" feature
Favouriting controller feature isn't updating redux store properly onClick resulting in inconsistent view in frontend.

1. Need to make sure favourite  feature uses unique result/controller names

anishaswain:
Using unique record name as the filter instead of keys which aren't unique.
Fixes #188 
Depends on PR: #191


dbutenhof (dismissed):

src/models/user.js
```diff
@@ -73,13 +73,13 @@ export default {
     removeFavoriteController(state, { payload }) {
       return {
         ...state,
-        favoriteControllers: state.favoriteControllers.filter(item => item.key !== payload.key),
+        favoriteControllers: state.favoriteControllers.filter(item => item !== payload),
```
aquibbaig, line 76: @anishaswain rather than doing referential equality here can we use some function that does "deep equality" like lodash's isDeepEqual(). I am being a little paranoid for edge cases here and I may be wrong

anishaswain: @aquibbaig in every case the key will be a string( here item is the result name). We could have gone for a deep comparison if it was an object or array. But for string comparison, I guess shallow comparison should be enough.

aquibbaig: I didn't know that it was a string, sorry!

aquibbaig (dismissed):

PR #195, opened 2021-08-23
Fixes can't compare iterations

### LINKED ISSUE #180: can't compare iterations
Demo
URL: http://pbench.perf.lab.eng.bos.redhat.com/dashboard/#/dashboard/comparison

aquibbaig:
Note

- Compare this PR against b2.1 (production branch)
- This PR fixes #180


aquibbaig:
@npalaska @dbutenhof The query to ```request.post(`${endpoint}&scroll_id=${allData._scroll_id}`)``` still fails with 502, I am not sure how to debug this case, can someone help me with the proper request endpoint?

dbutenhof:
> @npalaska @dbutenhof The query to `request.post(``${endpoint}&scroll_id=${allData._scroll_id}``)` still fails with 502, I am not sure how to debug this case, can someone help me with the proper request endpoint?

What exactly are you doing? I assume this is in the service `scrollUntilEmpty` code; but what's `data` at this point? *Is* there, for example, a `_scroll_id`? Also, what Elasticsearch configuration are you using for this? This appears to be the "old" dashboard code, which is calling Elasticsearch directly rather than using the Pbench passthrough API; but I'm not really sure what this b2.1 branch has... is this before or after the conversion to Elasticsearch 7, for example?

aquibbaig:
@dbutenhof you are absolutely right, this is in the `scrollUntilEmpty` code.

The data parameter means the following I guess:
- for each selected iteration we create a request and add it to an array called `iterationSampleRequests` [here](https://github.com/aquibbaig/pbench-dashboard/blob/a34302ffe03610d029df2c67cd77f6ebf6c4acd9/src/services/dashboard.js#L172)
- then we wait for all the requests to be complete using Promise.all() which is [here](https://github.com/aquibbaig/pbench-dashboard/blob/a34302ffe03610d029df2c67cd77f6ebf6c4acd9/src/services/dashboard.js#L245)
- the response recieved for each iterationSampleRequest is the data here I think which is passed to `scrollUntilEmpty`

The error I am getting is a 502, which means that the response is an HTML containing text such as "the requested data does not exist" and from the response we are trying to extract the scroll id in [scrollUntilEmpty](https://github.com/aquibbaig/pbench-dashboard/blob/a34302ffe03610d029df2c67cd77f6ebf6c4acd9/src/services/dashboard.js#L14) which will get undefined of course.

I believe this code is before elasticsearch v7 and yes, it queries ES directly, I believe @portante will be able to shed some light on this.

src/models/dashboard.js
```diff
@@ -189,26 +189,30 @@ export default {
...
-        Object.entries(clusteredIterations[primaryMetric]).forEach(([clusterId, cluster]) => {
-          const clusterKey = `${runId}_${iterationName}_${sampleName}`;
-          if (clusterKey in cluster) {
-            clusteredIterations[primaryMetric][clusterId][
-              clusterKey
-            ].timeseries = timeseriesCollection;
-          }
-        });
+          Object.entries(clusteredIterations[primaryMetric]).forEach(([clusterId, cluster]) => {
+            const clusterKey = `${runId}_${iterationName}_${sampleName}`;
```
dbutenhof: All of the variables referenced here were pulled out of `hits.hits[0]` and certainly don't change based on the iteration through the `clusteredIterations` array that's passed in, so it could be pulled out of the loop.

dbutenhof:
So you're saying that the original Elasticsearch query isn't just returning `hits.hits[]` but an error, which isn't being recognized before passing it to `scrollUntilEmpty`? Why is that? And maybe we need to be checking whether we have the `hits.hits` payload at all before trying to check the length ... as well as possibly checking the Elasticsearch HTTP status first?

I still really don't have a clear picture of what's happening here. I don't know the payload, I don't fully understand the fairly complex processing being done in the loops which are issuing multiple Elasticsearch queries for each element in the props and then trying to stitch together the `Promise.all` results through the `map` operation. (Does the `map` even really work, since it seems to be assigning to a local variable that was a copy of the reference to a result element?)

aquibbaig:
@dbutenhof yes, it is returning an error, maybe the endpoint is incorrect?

I have edited the post-processing to function correctly at the moment in this PR but the request itself is failing

dbutenhof (commented):
> @dbutenhof yes, it is returning an error, maybe the endpoint is incorrect?
> 
> I have edited the post-processing to function correctly at the moment in this PR but the request itself is failing

What's the final endpoint? Looks like `(indices)/_search`  but that leaves a lot of unknowns depending on the date range in the props, along with the index prefix and name from from the config file. Knowing the full text of the error return document might help, too.

src/models/dashboard.js
aquibbaig: @dbutenhof I'm sorry, I meant the request may be wrong, the request to the ES ep is ```request.post(``${endpoint}&scroll_id=${allData._scroll_id}``)```, which returns a 502

dbutenhof: Sorry; I didn't mean to suggest this comment has anything to do with your problem.

dbutenhof:
Again, knowing the actual Elasticsearch URL that's causing a 502 (along with the full response document) might help.

I did a search for Elasticsearch returning 502 errors, and one article I hit suggested they'd seen a 502 for low storage in the Elasticsearch instance. I'm not sure how Elasticsearch handles a scroll context, but I wonder if it's possible that your initial `&scroll=1m` query is trying to create a scroll document with so much data that it's running out of temporary memory? On the other hand, if you're really seeing this on the follow-up `&scroll_id=id`, presumably the storage has already been allocated and we're just reading it.

aquibbaig:
@dbutenhof Thanks for looking into this problem for me, temporarily I have used a fallback to "not show the time-series data for the iteration if the request fails", I think that fixes our problem in some way, if it is from elasticsearch's end, I do not know where to look at. Also, I have added a spinner to let users know that data is being fetched.

src/models/dashboard.js
dbutenhof: I'll try again: my point was that it seems that line 208 could be moved outside the loop at line 207 because the variables it references won't change across iterations of the loop.

src/models/dashboard.js
```diff
@@ -189,26 +189,30 @@ export default {
...
-        const primaryMetric = firstResponse.sample.measurement_title;
-        const iterationName = firstResponse.iteration.name;
-        const sampleName = firstResponse.sample.name;
-        timeseriesResponse.hits.hits.forEach(timeseries => {
-          timeseriesCollection.push({
-            x: timeseries._source['@timestamp_original'],
-            [`y-${runId}_${iterationName}_${sampleName}`]: timeseries._source.result.value,
+        // gets firstResponse only when there is a
+        // non empty array of hits.
+        if (timeseriesResponse.hits.hits.length > 0) {
```
dbutenhof: Do we want to consider being even more "paranoid" here, not assuming that we necessarily even have a "standard" Elasticsearch success payload here? This will fail, for example, if the payload doesn't have "hits" because what we got was an Elasticsearch error message document. That seems to be what you said you were seeing, but this code doesn't address that possibility.

src/services/dashboard.js
```diff
@@ -1,22 +1,43 @@
...
 import { getAllMonthsWithinRange } from '../utils/moment_constants';
 
 async function scrollUntilEmpty(datastoreConfig, data) {
   const endpoint = `${datastoreConfig.elasticsearch}/_search/scroll?scroll=1m`;
   const allData = data;
-
   if (allData.hits.total !== allData.hits.hits.length) {
-    const scroll = Promise.resolve(request.post(`${endpoint}&scroll_id=${allData._scroll_id}`));
+    const scroll = Promise.resolve(
+      request.post(`${endpoint}&scroll_id=${allData._scroll_id}`, {
```
dbutenhof, line 13: Ah; I wasn't entirely sure before, but I see that you were getting the 502 on the follow-up scroll queries rather than on the initial query. That makes even less sense to me, and I wish we had more context to figure out what's going on with Elasticsearch here. I'm still concerned that most of our code isn't protected against unexpected Elasticsearch errors; for example, if we got the 502 on the initial request rather than on the scroll queries.

Then again, this is maintenance on the old dashboard code, so limiting the change is a good idea. What can we do on `main` to make our dashboard more robust in handling back-end and connectivity errors?

dbutenhof (dismissed):

src/services/dashboard.js
aquibbaig: I believe we have error handling work #185 in the `main` branch that catches errors like these and displays them to the user.

src/models/dashboard.js
aquibbaig: Let me add the conditions for extra safety.

dbutenhof (approved):

anishaswain (approved):
LGTM!

portante (approved):

PR #197, opened 2021-09-02
modify jest transformIgnorePatterns to fix es6 exports

aquibbaig:
aimed at fixing unit tests


aquibbaig:
Conclusion:
- It somehow makes sense to jest to give values with `?!` prefix in a single line separated by `|` operator, else it doesn't individually transform it. (this is a completely experimental result I found)

aquibbaig:
Note: I could not find many resources on how to modify babel in a `umijs` environment which could have fixed the issue, but the fix I have here is purely experimental and seems to be consistent across my local environment and GitHub Actions

portante (approved):
Let's just open an issue to track a long-term fix for this going forward.

dbutenhof (approved):
Is there some process for phasing in new ECMAScript support to decide when we no longer transform packages for testing? This should all be captured in the issue for improving this in the future. Given the diversity of Javascript packages maybe the basic requirement for this stuff will never go away, but it seems like a permanent weakness. (Though it seems everyone's just used to dealing with it.)

aquibbaig:
@dbutenhof You're absolutely right, in normal cases a `.babelrc` configuration for babel would transpile these files, but I do not know how we integrate this with umi

PR #199, opened 2021-09-07
Migrate DatePicker component from Ant Design 

### LINKED ISSUE #182: Migrate DatePicker component from Ant Design

anishaswain:
Fixes #182


portante (dismissed):
Looks good!

portante (dismissed):

src/components/DatePicker/index.js
```diff
@@ -1,82 +1,63 @@
...
-class AntdDatePicker extends Component {
-  static propTypes = {
-    onChangeCallback: PropTypes.func,
-  };
+const PFDatePicker = props => {
+  const [from, setFrom] = useState('');
+  const [to, setTo] = useState('');
+
+  const toValidator = date =>
+    isValidDate(from) && date >= from ? '' : 'To date must be less than from date';
```
dbutenhof: Is this right? It seems that `date` here is a potential `to` date, being compared against `from`, but the error on failure is that "To date must be less than from date". Isn't that backwards? We're failing if `date` is `< from` because we expect `to` date to be the same or later than `from`.

dbutenhof (commented):
Cool; this looks a lot simpler. I'm not sure why this has to be `PFDatePicker` instead of just `DatePicker`, but that's a technicality.

I do think there's a mistake in the validator message, however.

src/components/DatePicker/index.js
anishaswain: @dbutenhof the reason behind using `PFDatePicker` is , Patternfly is already using the keyword `DatePicker` for its component and we are also using the same in our page. So there would be a clash of same keywords if we don't use separate names.

anishaswain: Fixed the `to` , `from` issue.

anishaswain:
@portante @dbutenhof @aquibbaig please re-review.

aquibbaig (commented):
https://user-images.githubusercontent.com/26324376/138043007-ddfa2830-85c3-483f-8526-4c26a6f83266.mov

Wondering why it resets the datepicker when I pick the `to` case?

src/components/DatePicker/index.js
```diff
@@ -1,82 +1,63 @@
...
-  static defaultProps = {
-    onChangeCallback: () => {},
+  const onFromChange = (_str, date) => {
+    setFrom(date);
   };
 
-  handleChange = (dates, dateStrings) => {
-    const { dispatch, onChangeCallback } = this.props;
+  const getDataOnChange = (fromDate, toDate) => {
+    const { dispatch, onChangeCallback } = props;
```
aquibbaig: this line can move to the top of the function component scope

src/components/DatePicker/index.js
```diff
@@ -1,82 +1,63 @@
...
+import { Split, SplitItem, DatePicker, isValidDate, yyyyMMddFormat } from '@patternfly/react-core';
+import React, { useState } from 'react';
 import { connect } from 'dva';
-import moment from 'moment';
-import PropTypes from 'prop-types';
-import './index.less';
 
-const { RangePicker } = DatePicker;
+const mapStateToProps = state => {
+  return state;
```
aquibbaig: Can we destructure what's necessary for this component so that we have a little bit of idea on what this component receives as props

src/components/DatePicker/index.js
anishaswain: fixed!

src/components/DatePicker/index.js
anishaswain: fixed!

anishaswain:
> Wondering why it resets the datepicker when I pick the `to` case?

Fixed!

src/e2e/controllers.e2e.js
```diff
@@ -120,6 +120,7 @@ describe('controller page component', () => {
   });
 
   test('should sort last modified column chronologically ascending', async () => {
+    await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] });
```
aquibbaig, line 123: @anishaswain What does this change do?

anishaswain: It refreshes the page before any other page click happens. Seems like e2e tests weren't being able to find the same button after once clicked.

aquibbaig: ok

aquibbaig (dismissed):

src/components/DatePicker/index.js
```diff
@@ -11,7 +11,7 @@ const PFDatePicker = props => {
   const [to, setTo] = useState('');
 
   const toValidator = date =>
-    isValidDate(from) && date >= from ? '' : 'To date must be less than from date';
+    isValidDate(from) && date >= from ? '' : '"From" date must be less than or equal to "To" date';
```
dbutenhof: This phrasing sounds a bit odd; you're validating the "to" date but the sentence focuses on the "From" date. Instead, it would seem more clear to say that the "to" date must be greater than or equal to the "from" date.

However, the statement isn't "incorrect", it just sounds odd, so I'll approve and let you decide whether to make another change. 😁

dbutenhof (dismissed):

src/components/DatePicker/index.js
anishaswain: fixed!

portante (approved):

dbutenhof (approved):
Thanks!

PR #204, opened 2021-10-21
fixes e2e test consistent results on controllers page

### LINKED ISSUE #203: e2e tests cant find the same element without doing a page reload
```
await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] });
```

is hacky fix to move around the issue, we need to get an "actual fix"

aquibbaig:
Earlier, for some reason, we were searching the controller's name in the datepicker's input (weird). This PR also removes chronological sorting of `last_modified` columns  because these are constant across all the records (sorting it causes inconsistent behaviour at times)

Fixes #203


aquibbaig:
We can "add different values of last_modified in the mock data"

portante (changes requested):
Let's turn this into multiple commits:

 1. One that fixes the existing tests without dropping the 2 tests that are sorting
 2. One that replaces the sorting tests with working sorts

aquibbaig:
Changes in this PR
- Uses proper sorting methods (for example: comparing last_modified values in the table with mock values) rather than alphabetical matching
- resets table after each sorting rule (There are 3 sorting rules present in react table as per my experience: ascending, descending and default, so after you view the first two, clicking on the header once more resets the table into default so that you can test other sorting rules from the start)
- we generate new values of last_modified using the moment library

webbnh (approved):
Looks good as far as I can tell.

dbutenhof (approved):
I find most of this mysterious without trying to trace through the DOM, but I'm assuming this is probably good. :-)

PR #206, opened 2021-11-08
Working endpoints

### LINKED ISSUE #193: Search component not rendering table data
In the mock environment, when we hit the search button, no items show up on the table component

### LINKED ISSUE #194: can't fetch from /api/v1 to fetch endpoints
For testing with a real server, we need endpoints served at `/api/v1` to populate endpoints in the dashboard.

aquibbaig:
This PR is a `umijs` way of managing endpoints in a React app, aimed at fixing #194 

This PR does the following
- references `window.endpoints` directly in services/ folder. Declaration of `const { endpoints } = window` on a top level for each file works at runtime and does not get updated when the endpoints are reassigned.
- endpoints are reassigned in BasicLayout(top level HOC in our app tree) by querying the actual server endpoint which populates the browser with all server endpoints accessible via `window.endpoints`. **Note: This URL is intercepted by mock server in dev environment**
- updates runtime configuration for prod and dev cases. In ideal cases, you have `config.js` for dev environment, `config.prod.js` for production & `config.staging.js` for staging

TODO:
- [x] Proper naming of the pbench server variable in runtime config. Currently, it is named as `ENDPOINT`.
- [ ] Updating documentation for the new configuration system


aquibbaig:
**Tests to be resolved after #208**

src/layouts/BasicLayout.js
```diff
@@ -64,6 +64,21 @@ class BasicLayout extends React.PureComponent {
     // find the current route from menuData.
     const currRouteMap = this.breadcrumbNameMap.find(route => route.path.endsWith(pathname));
 
+    if (!window.endpoints) {
+      // eslint-disable-next-line no-undef
+      fetch(`${ENDPOINT}`).then(response => {
```
dbutenhof: Why `${ENDPOINT}` instead of just `ENDPOINT`?

config/example.config.js
```diff
@@ -0,0 +1,19 @@
+import pageRoutes from './router.config';
+
+export default {
+  // Note: typeof(process.env.MOCK) === string
+  define: {
+    MOCK_UI: process.env.MOCK === 'true',
+    // Use http://localhost:8000/api/v1/endpoints for development environment
+    // to reference the mock dev server.
+    ENDPOINT: 'PBENCH_SERVER_ENDPOINT',
```
dbutenhof: I don't understand the intent of this comment. In real production we want to fetch from `/api/v1/endpoints` to implicitly use `window.origin` host; for your "hosted developer mode" you want to point to a real server like `http://pbench-server.example.com/api/v1/endpoints`; you don't ever really want to point to your local npm server. In the local e2e, we *mock* the `/api/v1/endpoints` call entirely without changing the `fetch` target.

webbnh (commented):
I'm confused about the config files:  the `.gitignore` change seems to anticipate a `config/config.prod.js` and a `config/config.staging.js`, but the PR includes a `config/config.js`.  (Are the first two "fake"/temporary files, while the last is the "real" file?). And, you're also committing `config/example.config.js` which would seem to be the template that the preceding three are supposed to be constructed from.  But, if we want to commit the template, then perhaps we do _not_ want to be committing the "live" one (i.e., should `config/config.js` be included in the PR, or should it be ignored like the other two `config/config*.js` files?)....

src/services/search.js
```diff
@@ -2,16 +2,12 @@
...
 export async function queryIndexMapping(params) {
   const { indices } = params;
 
-  const mappings = `${endpoints.indices.run_index}${indices[0]}/_mappings`;
+  const mappings = `${window.endpoints.indices.run_index}${indices[0]}/_mappings`;
 
-  const endpoint = MOCK_UI
-    ? endpoints.api.mappings
-    : endpoints.api.elasticsearch;
+  const endpoint = MOCK_UI ? endpoints.api.mappings : endpoints.api.elasticsearch;
```
webbnh: Shouldn't you have `endpoints` prefixed with `window.` in this case?

webbnh (dismissed):
As far as I can tell, this looks OK, except for the one item that I questioned below, and the question about the config files above.

And, possibly the failing tests....

src/services/auth.js
```diff
@@ -1,9 +1,7 @@
 import request from '../utils/request';
 
-const { endpoints } = window;
```
dbutenhof, line 3: I'm curious why you removed the local unpacking of `endpoints` from `window` throughout; this isn't really a functional change, and seems to just complicate the PR with non-essential edits. (I don't know whether the destructing actually *copies* `window.endpoints` to the local `endpoints` or just constructs a reference: in the former case, your change might be a bit more efficient... but I would hope that it's just a mapped reference without moving any data.)

config/example.config.js
aquibbaig: This file needs to go from this PR, the development configuration is `config/config.js` by default. For prod, you can use `config/config.prod.js`

src/services/auth.js
aquibbaig: Yeah, it has to do with the `closure` property of js language. On initial load of `umi.js` script, the variable `window.endpoints` is null. However with due progression of the dashboard, we reassign this variable all of our endpoints fetched from the server.

However, unpacking it at the global level doesn't point to this reassigned `window.endpoints` and is perennially `null`.

aquibbaig: You were earlier writing a custom script above `umi.js` script in `document.ejs` and assigning the endpoints synchronously earlier in your endpoints work and that's why this system was working for you

aquibbaig:
@webbnh I have used a sample configuration from https://v2.umijs.org/guide/env-variables.html#umi-env. 
I will delete `example.config.js` because for the local environment a simple `config.js` works fine

aquibbaig:
FYI, the last commit in this PR fixes #193

src/services/dashboard.js
```diff
@@ -111,7 +109,7 @@ export async function queryIterationSamples(params) {
     selectedDateRange
   )}/_search?scroll=1m`;
 
-  const endpoint = MOCK_UI ? endpoints.api.datasets_samples : endpoints.api.elasticsearch;
+  const endpoint = MOCK_UI ? window.endpoints.api.datasets_samples : endpoints.api.elasticsearch;
```
dbutenhof: Looks like you missed adding `window.` on the "else" case here?

src/layouts/BasicLayout.js
```diff
@@ -64,6 +64,27 @@ class BasicLayout extends React.PureComponent {
     // find the current route from menuData.
     const currRouteMap = this.breadcrumbNameMap.find(route => route.path.endsWith(pathname));
 
+    // eslint-disable-next-line no-undef
+    fetch(ENDPOINT)
```
dbutenhof, line 68: I still don't understand what you're doing with `ENDPOINT`. Your config change has it pointing at `localhost:8000` which I wouldn't expect will ever respond. For real production deployments we want to fetch from `/api/v1/endpoints` (implicitly `window.origin`) while for your local development mode you want to point to the real endpoint of an actual Pbench server, like `http://10.1.170.1/api/v1/endpoints`. You're not doing either. Are you intending to build an npm "reflector" that does handle `http://localhost:8000/api/v1/endpoints` by calling a real `http://10.1.170.1/api/v1/endpoints` (as we'd discussed once before) and returning the result? And, if so, where is that code and where is the actual remote address defined?

The code as presented in this PR will break both modes, so I'm still confused about where you plan to take this.

aquibbaig: Yeah, in the language of umijs, the variables that you **define** in `config/config.js` are publicly available in every js file. So this `ENDPOINT` variable is defined in the config file. When you wish to connect to the real server, you replace this `ENDPOINT` variable in the `config/config.js` to `http://10.1.170.1`.

dbutenhof: The code in the repo should be configured for real deployment; and in this case, the real deployment target should be what was in `document.ejs` before: simply `/api/v1/endpoints`, to reference the server API using the default `window.origin` path. We don't want to go back to requiring manual dashboard configuration for product deployment.

For local developer mode, it's acceptable to have to edit the local config and manually supply the target server; but I'd also like to make sure that such changes don't accidentally get merged back into the repo, so ideally it would be done with a separate file that can be listed in `.gitignore` rather than editing the real config file.

And I still don't understand why you've got `http://localhost:8000/api/v1/endpoint` as the default, which shouldn't do anything at all unless you've got a local server responding to that request ... which isn't in this PR at all.

webbnh: I'll second what Dave said:  as committed to the repo, the code should be production-ready.

However, as Aquib [pointed out above](https://github.com/distributed-system-analysis/pbench-dashboard/pull/206#issuecomment-964009418), you can use [`UMI_ENV`](https://v2.umijs.org/guide/env-variables.html#umi-env) to seamlessly overlay the default (production) configuration with a local/developer configuration (and, the `.gitignore` file in this PR will decline to commit those configuration files).

So I think we're close to what we want, we just need to set the to-be-committed `config/config.js` to reflect what the production configuration should be.

webbnh (dismissed):
Other than the default configuration being deployment-ready, I've got no comments.

src/layouts/BasicLayout.js
aquibbaig: We cannot use `window` object in a nodejs build such as `config/config.js`. It will always be undefined and wont start the yarn web server.

aquibbaig:
Note: For auto deploying the dashboard we would need to use `window.origin` somewhere. We cant use the window object in `config/config.js` since that is a nodejs context. 

The other way around is to set the ENDPOINT variable as `ENDPOINT: /api/v/endpoints` and prefix our requests in `services/` directory with `window.origin` instead.

src/layouts/BasicLayout.js
dbutenhof: Sure; but `ENDPOINT = /api/v1/endpoints` will work just fine. It's only for local developer mode that there's a need to override the default `window.origin` host/port.

aquibbaig:
The future of this PR depends upon how we would want to go forward with our frontend solution. If we want a version of the dashboard that is auto configured when the server is deployed, I think the "templating" solutions provided by Flask or Django that render a view from the server side can be more suitable rather than an independent React app.

This PR however makes debugging on the local side with a real server easier.

src/services/auth.js
```diff
@@ -21,7 +19,7 @@ export const postRegisterUser = async params => {
 };
 
 export const postLoginUser = params => {
-  const endpoint = `${endpoints.api.login}`;
+  const endpoint = `${window.origin}${window.endpoints.api.login}`;
```
dbutenhof: Why `${window.origin}${window.endpoints.api.login}`?? This is wrong, since the value returned for `api.login` from the endpoints API will be a fully formed URI like `http://10.1.170.201/api/v1/login`.

You also seem to have added the `window.origin` prefix in just a few of the services; all of those will be broken against a real server.

dbutenhof (changes requested):

config/config.js
```diff
@@ -4,10 +4,11 @@ export default {
   // Note: typeof(process.env.MOCK) === string
   define: {
     MOCK_UI: process.env.MOCK === 'true',
+    ENDPOINT: 'http://localhost:8000/api/v1/endpoints',
```
dbutenhof, line 7: Now this is broken again! We shouldn't be changing the "real" `config.js` for local testing. I'd assumed that was the point of your `.gitignore` changes, where you'd have alternate unstaged config files *without* changing the real version. Someone's going to end up accidentally committing this (as you've already done) and breaking the dashboard; we really don't want that to happen.

src/services/global.js
```diff
@@ -1,10 +1,8 @@
 import request from '../utils/request';
 
-const { endpoints } = window;
-
 export async function saveUserSession(params) {
   const { sessionConfig, description } = params;
-  const endpoint = `${endpoints.api.graphql}`;
+  const endpoint = `${window.endpoints.api.graphql}`;
```
dbutenhof, line 5: You don't need the backtick string substitution here when `const endpoint = window.endpoints.api.graphql` would work just as well. (And in fact, the `endpoint` local really doesn't help much anymore; it's already been removed in most other paths.)

src/services/global.js
```diff
@@ -26,7 +24,7 @@ export async function saveUserSession(params) {
 
 export async function queryUserSession(params) {
   const { id } = params;
-  const endpoint = `${endpoints.api.graphql}`;
+  const endpoint = `${window.endpoints.api.graphql}`;
```
dbutenhof, line 27: Same here

src/services/sessions.js
```diff
@@ -48,7 +46,7 @@ export async function getAllSessions() {
 
 export async function saveSession(params) {
   const { sessionConfig, description } = params;
-  const endpoint = `${endpoints.api.graphql}`;
+  const endpoint = `${window.endpoints.api.graphql}`;
```
dbutenhof, line 49: And here, and several additional places in the file. It's not incorrect; but it'd look neater without the backtick substitution.

dbutenhof (changes requested):
We can't have a commit with `ENDPOINT` set to anything but the normal production value of `/api/v1/endpoints`!

PR #208, opened 2021-11-08
refetch search results after selection of dates from date picker

### LINKED ISSUE #207: Search page throws error when filtered using date picker
Currently our Search page throws the following error when we filter search results using date picker component
```
TypeError: onChangeCallback is not a function
```

This happens because the datepicker expects a callback function to be executed after the date selection, currently we do not pass any in the Search page.

aquibbaig:
fixes #207


dbutenhof (dismissed):
Plus, apparently, switching the search pane from Antd to Patternfly date picker. A bit more of a git log description would be good, rather than just a reference to the issue; but the change seems good.

webbnh (approved):
Looks OK to me.  (Not sure that that is saying much, but....)

npalaska (approved):
looks good, Its a small change and if it fixes the issue, GTG. :)

PR #212, opened 2021-12-10
Add support for creating an RPM for the Dashboard

### LINKED ISSUE #186: Ensure the pbench-dashboard code is deployable repeatedly with the pbench-server
Ensure the pbench-dashboard code is deployable repeatedly with the pbench-server, as by a functional test and a naive developer.

The goal is to verify behavior and generate issues as needed to cover missing functionality or behaviors.

webbnh:
The recent [effort to containerize the Pbench Server deployment](https://github.com/distributed-system-analysis/pbench/pull/2585) includes deploying a Dashboard with the Server in the container.  This is done by building an RPM for the Dashboard and installing it in the container as a dependency of the Server installation.  To support that, this PR provides the facility for building an RPM for the Dashboard.


rpm/Makefile
```diff
@@ -0,0 +1,100 @@
...
+# 3. Update any GIT submodules.
+# 4. Do a "make install" to a temp directory.
+# 5. Generate a tar ball from the directory.
+# 6. Generate a local SRPM that will be uploaded to COPR for building.
+# 7. Optionally generate a local RPM.
+# 8. Clean up the temp directory
+
+# NOTE:
+# Currently, we have a static spec file instead of generating one from a
+# template, so the version number, etc. is not dynamically generated.
```
dbutenhof: Sure you don't want to fix this before it goes in? 😁

dbutenhof (dismissed):

rpm/Makefile
webbnh: I think this is a "perfect is the enemy of the done" thing -- once we decide that we're going to carry the existing Dashboard forward, perhaps I'll be more enthusiastic about addressing this.  🙂

rpm/Makefile
```diff
@@ -0,0 +1,100 @@
...
+	rm -rf ${TMPDIR}/*
+
+.PHONY: rpm-dirs
+rpm-dirs:
+	mkdir -p $(addprefix ${HOME}/rpmbuild/,${RPMDIRS})
+
+.PHONY: submodules
+submodules:
+	git submodule update --init --recursive
+
```
ndokos: Does the dashboard use submodules? If not, this (plus any references to it) could be deleted.

ndokos (commented):
LGTM. I'll wait on the answer to the question before approving.

rpm/Makefile
```diff
@@ -0,0 +1,100 @@
...
+#	jinja2 ${prog}.spec.j2 -D version=${VERSION} -D gdist=g${sha1} -D seqno=${seqno} > ${RPMSPEC}/${prog}.spec
+#	rpmlint --file ${PBENCHTOP}/utils/rpmlint ${RPMSPEC}/${prog}.spec
+
+.PHONY: patches
+patches: rpm-dirs
+	if [ -d ./patches ] ;then cp ./patches/* ${RPMSRC}/ ;fi
+
+.PHONY: tarball
+tarball: rpm-dirs submodules ${subcomps}
+	echo "${sha1}" > ${TBDIR}/${component}/SHA1
```
ndokos: Does `${TBDIR}/${component}` need to be created?

rpm/Makefile
```diff
@@ -0,0 +1,100 @@
...
+
+component = dashboard
+subcomps =
+
+
+PBENCHTOP := $(shell git rev-parse --show-toplevel)
+TMPDIR = /tmp/opt
+
+prog = pbench-${component}
+VERSION := $(file < ${PBENCHTOP}/${component}/VERSION)
```
ndokos: AFAICT, there is no VERSION file - correct? At least, not under `${PBENCHTOP}/${component}/VERSION`.

rpm/Makefile
webbnh: I just checked, and the Dashboard does not current have any submodules.  However, since pretending that it does have some doesn't hurt anything, and having been burned by failing to be mindful of them in the past, I'm inclined to retain this.

That said, I'm willing to remove it if you feel strongly about it.

ndokos:
`yarn` is provided by `yarnpkg` (on F34 at least), so the dependency in the spec needs to be modified.

webbnh:
[Apparently](https://docs.pagure.org/copr.copr/user_documentation.html#:~:text=Note%20that%20starting%20from%20December%202021%2C%20Copr%20migrated%20to%20the%20rpkg%2Dutil%20v3) COPR migrated to `rpkg-util v3` in December 2021 (while I was on PTO), so the spec file now needs some tweaking, so I'm putting this PR into Draft mode while I get things working again.

(Also, I suspect that I don't even need the `Makefile`, so that may be going away.)

webbnh:
When using COPR to build an RPM from a Git repository, it only needs to know how to access the repo and branch and where to find the spec file, so the `Makefile` which I previously included in this PR is unnecessary (so I've removed it).  COPR (via `rpkg-util`) is clever enough to add build numbers, Git hashes, et al. to the package name (so the NVR format is properly managed), and so a lot of what we were doing in the `Makefile` is handled automagically.

I moved the spec file from the `rpm` subdirectory to the top level.  By default, `rpkg-util` builds the source tarball starting at the directory where the spec file is, so, rather than try to persuade it to do otherwise, I just moved the spec file to where it wants to find it (which is fine -- since we now have no other files that we would put in an `rpm` subdirectory, there's little point in keeping it there).

With those changes, the RPM build seems to be working again, so this PR is once again ready for review.

webbnh:
> `yarn` is provided by `yarnpkg` (on F34 at least), so the dependency in the spec needs to be modified.

@ndokos, in the settings for the "project" on COPR, there is a place to enter "External Repositories"; I've included `https://dl.yarnpkg.com/rpm/` there, which seems to make the `BuildRequires` reference to `yarn` work (at least on `epel-8-x86_64` -- I didn't need it on F34, so I didn't try that chroot).

rpm/Makefile
webbnh: Given that I've removed this file from the PR, presumably, this is now moot.

rpm/Makefile
webbnh: I think you are correct, but, given that I've removed this file from the PR, presumably, this is now moot.

rpm/Makefile
webbnh: I think you're correct, but, given that I've removed this file from the PR, presumably, this is now moot.  Also, this functionality is provided for by `rpkg-util`, so we should now be covered.

rpm/Makefile
webbnh: FWIW, this functionality is provided for by `rpkg-util` which COPR uses to build the RPM, so we should now be covered.  And, given that I've dropped this file from the PR, I think it's now moot.

pbench-dashboard.spec
```diff
@@ -0,0 +1,46 @@
...
+yarn install
+yarn build
+
+%pre
+getent group pbench > /dev/null || groupadd pbench
+getent passwd pbench > /dev/null || useradd -g pbench -d /home/pbench -c "Pbench user" pbench
+
+%install
+rm -rf %{buildroot}
+mkdir -p %{buildroot}/%{installdir}
```
dbutenhof, line 33: Is it really necessary to do the `rm -rf` here? I worry about doing that with any "unknown" parameter, in general. If you really feel the need to make sure that `%{installdir}` is initially empty, `rm -rf %{buildroot}/%{installdir}` would seem safer since we know that's something we're going to own entirely regardless of the root.

dbutenhof (approved):
I'll approve, despite a comment; because the server build spec file also does the `rm -rf %{buildroot}` and it's presumably reasonable to assume that `%{buildroot}` won't be something like `/` or `${HOME}`; but it still seems like an awkward idiom to me. ☹️

pbench-dashboard.spec
webbnh: My understanding is that the RPM build is (i.e., in the general case) performed in a temporary/restricted environment (e.g., a "chroot"), so the massive `rm` is actually a safe and reasonable thing to do.

webbnh: Your concerns, are valid -- apparently the practice is widespread/standard enough that there is [a warning around using the `--buildroot` switch](http://ftp.rpm.org/max-rpm/ch-rpm-b-command.html#S3-RPM-B-COMMAND-BUILDROOT-WARNING).

(BTW, can you resolve your previous "conversation"?)

portante (approved):

ndokos:
> in the settings for the "project" on COPR, there is a place to enter "External Repositories"; I've included https://dl.yarnpkg.com/rpm/ there, which seems to make the BuildRequires reference to yarn work

I'm not convinced this is a good idea: what happens if I want to build an RPM from your branch in my COPR repo? I have to have the setting as well. Also, don't we want the distro-packaged RPM, rather than whatever version is packaged at `dl.yarnpkg.com`?

EDIT: I guess one problem that the above solves is the unavailability of a  `yarnpkg` RPM on RHEL8 - sigh...

ndokos (approved):

webbnh:
@ndokos [wrote](https://github.com/distributed-system-analysis/pbench-dashboard/pull/212#issuecomment-1005219496):
> > in the settings for the "project" on COPR, there is a place to enter "External Repositories"; I've included https://dl.yarnpkg.com/rpm/ there, which seems to make the BuildRequires reference to yarn work
> 
> I'm not convinced this is a good idea: what happens if I want to build an RPM from your branch in my COPR repo? I have to have the setting as well. Also, don't we want the distro-packaged RPM, rather than whatever version is packaged at `dl.yarnpkg.com`?

There are two sets of dependencies:  install-time and build-time -- install-time dependencies are satisfied from the host where the RPM is installed, while build-time dependencies are used where the RPM is _built_.  If we depended on `yarn` at install-time, I would agree that we would want/need a distro-packaged RPM for it.  However, since the `yarn` dependency is a build-time dependency, what is needed is to tell COPR how to install it on the VM used to build the RPM.

The fact that our dependency on `yarn` is satisfied at build-time instead of at install-time trades one set of issues for another.  For instance, once our RPM is built, the set of JS dependencies that we select is frozen, which means that our users won't run into fresh incompatibilities when our RPM is installed; on the other hand, if a JS component is fixed after we build our RPM, we won't pick up that fix until we respin our RPM.  It's not obvious which of these is "better", but my instinct is that running `yarn` at RPM build-time will produce fewer surprises and gives us more control over the results.

So, if you want to build a Dashboard RPM on COPR (i.e., using the web GUI), you have to set up a "project" and request a build.  (I don't know whether you can request builds in my project, but you can certainly set up a project under your own login.)  When you request the build, you can specify any publicly-accessible repo and branch, including one from my fork of the Dashboard repo.  And, you have to tell COPR how to satisfy the build-time dependencies...we currently have two -- `yarn` and `nodejs` -- apparently COPR knows how to get `nodejs`, but we have to tell it how to get `yarn`.  (See COPR details under "Build a Dashboard RPM" in [this document](https://docs.google.com/document/d/1k4zrnGNa8BdjcVvMWmFKpp26wGnXHpMJq6EA8kXOsnk).)

